PR #2, opened 2016-07-19
CLI Baseline Initial Commit

Tomdye:


Gruntfile.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function (grunt) {
+	require('grunt-dojo2').initConfig(grunt, {
+		intern: {
+			options: {
+				nodeOptions: [
+					'--harmony',
```
agubler: Do we still need the `--harmony` flags?

package.json
```diff
@@ -0,0 +1,61 @@
...
+  "name": "dojo-cli",
+  "version": "0.0.3-pre",
+  "description": "Dojo 2 CLI utility",
+  "homepage": "http://dojotoolkit.org",
+  "bugs": {
+    "url": "https://github.com/dojo/cli/issues"
+  },
+  "license": "BSD-3-Clause",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
agubler: we have some tests right?

Gruntfile.js
Tomdye: with Node 6.3 I still needed these two I believe. It complained that the `destructuring` one no longer existed but was happy about these.

package.json
Tomdye: writing them now

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "postinstall": "typings install"
```
agubler: Why is this `postInstall` and not in the build pipeline as per other modules?

Tomdye: this was before the other changes, if you point me in the right direction i will change

src/commands/newApp.ts
```diff
@@ -0,0 +1,267 @@
...
+	const renderResponses = await Promise.all(renderPromises);
+	const renderDisplayCount = chalk.green(renderResponses.length.toString());
+
+	log('info', chalk.bold.green('Summary: ') + `Rendered: ${renderDisplayCount}`);
+};
+
+function getSelectedModuleConfig(selectedModuleIds: string[], availableModuleConfig: ModuleConfigMap): ModuleConfigMap {
+	let modules: ModuleConfigMap = {};
+
+	// Get just the module config we care about
```
agubler: do we need to have a comment here?

Tomdye: no

Gruntfile.js
novemberborn: [Default parameters should be supported](http://node.green/#default-function-parameters).

Is there anything specific that `--harmony` enables? Would be good to enable just those.

Issue #10, opened 2016-08-11
help command


rishson:
Implement the `help` command.

This command should aggregate all the available commands and output a list of available commands:

```
dojo help

Hey there, here are all the things you can do with dojo-cli:

help - shows all the available commands
create - creates a new dojo application
run - ...

For more information on any of these commands just run them with '-h',
e.g. 'dojo run -h' will give you the help for the 'run' command. 

(You are running dojo-cli 2.3.20)
```

It shouldn't output the full help for every command as that will get too verbose. The top level `help` command is basically a command discovery point.

PR #16, opened 2016-08-11
Task discovery cli baseline

Tomdye:
# Work towards issue: #10 

Will display help including generated task descriptions when calling
```
dojo
```

## Example display
screenshot 2016-08-18 13 54 50

## Code coverage
screen shot 2016-08-25 at 16 01 17


src/lib/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function getCommandDescription(commandType: string, commandSubTypes: CommandConfig[]): string {
+	return commandSubTypes.length > 1 ?
+		`There are ${commandSubTypes.length} ${commandType} subCommands: ${commandSubTypes.map((commandSubType: CommandConfig) => commandSubType.name).join(', ')}` :
+		commandSubTypes[0].description;
+}
+
+function loadCommandFromPath(path: string): void {
+	const { description, register, run } = require(path);
+	const pluginParts = /dojo-cli-(.*)-(.*)/.exec(path);
```
matt-gadd: this should come from the `searchPrefixes` in the config no?

Tomdye: @matt-gadd would it not make more sense to remove `prefixes` from the config / make it a `string` rather than a `string[]` to be more opinionated about the `dojo-cli-` command naming conventions?

matt-gadd: @Tomdye i'd still leave it in the config (as it's used for the folder discovery at the moment), but maybe just change it to a string rather than an array of strings.

src/lib/config.ts
```diff
@@ -0,0 +1,15 @@
...
+	searchPrefixes: string[]
+};
+
+export default <Config> {
+	searchPaths: [
+		resolve(__dirname, '../node_modules'),
+		resolve(__dirname, '../../'),
+		'node_modules'
+	],
+	searchPrefixes: ['dojo-cli']
```
Tomdye: I think this should become `searchPrefix: 'dojo-cli'`

src/util/setupLogger.ts
```diff
@@ -0,0 +1,18 @@
+import { add, remove, transports } from 'winston';
+
+export interface VerboseOptions {
+	verbose?: boolean;
+}
+
+export default function setupLogger(verbose: boolean = false) {
```
matt-gadd: do we need this?

src/util/setupLogger.ts
```diff
@@ -0,0 +1,18 @@
...
+	remove(transports.Console);
+	add(transports.Console, {
+		showLevel: false,
+		level: verbose ? 'verbose' : 'info'
+	});
+
+	add(transports.File, {
+		filename: '.dojo-cli.log',
```
matt-gadd: i think we discussed this before, but this is a bad place to put the .log, especially in places where your commands demand an empty folder.

src/util/setupUpdateNotifier.ts
```diff
@@ -0,0 +1,6 @@
+import * as updateNotifier from 'update-notifier';
```
matt-gadd: can this file just be named updateNotifier? I don't know why we need to prefix with `setup` everywhere.

src/util/setupLogger.ts
agubler: Do we even want to have a `File` transport?

matt-gadd: I'm not even sure, I don't think this is in mvp anyway?

src/lib/index.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const helpEpilog = `For more information on any of these commands just run them with '-h'.
+
+e.g. 'dojo run -h' will give you the help for the 'run' command.
+
+(You are running dojo-cli ${pkg.version})`;
+
+function getCommandDescription(commandType: string, commandSubTypes: CommandConfig[]): string {
+	return commandSubTypes.length > 1 ?
+		`There are ${commandSubTypes.length} ${commandType} subCommands: ${commandSubTypes.map((commandSubType: CommandConfig) => commandSubType.name).join(', ')}` :
```
matt-gadd: we don't use camel case in plain english :P. how about just `commands` anyway?

src/util/setupLogger.ts
Tomdye: i'll take the file transport out then, users can use the verbose flag to see verbose output

matt-gadd:
it feels like the file structure to this is kind of just split up arbitrarily from the poc? I think it could be better organises logically, at the moment we seem to have files which serve one function and are named after it, yet others aren't?

src/util/setupLogger.ts
Tomdye: do we need to be able to enable verbose logging? Whats wrong with it?

src/lib/index.ts
```diff
@@ -0,0 +1,109 @@
...
+		commandSubTypes[0].description;
+}
+
+const commandRegExp = new RegExp(`${config.searchPrefix}-(.*)-(.*)`);
+function loadCommandFromPath(path: string): void {
```
matt-gadd: given all these functions have the word `command` in, perhaps create a `command` module and these can then all be nicely named `load`, `register`, `getDescription` etc

Tomdye: I thought about that, they'd all need to be passed the `taskMap` and `taskSet` though

matt-gadd: maybe evaluate how you construct this then.

src/util/setupLogger.ts
matt-gadd: like how useful is the logging in the mvp?

Tomdye: it's already there and it's small so I don't see the problem

matt-gadd: okay, well shall we remove the File transport and remove the `setup` from the filename

src/lib/index.ts
```diff
@@ -0,0 +1,109 @@
...
+import setupLogger, { VerboseOptions } from '../util/setupLogger';
+import setupUpdateNotifier from '../util/setupUpdateNotifier';
+import getGlobPaths from '../util/getGlobPaths';
+import * as chalk from 'chalk';
+import * as globby from 'globby';
+import config from './config';
+import { log } from 'winston';
+const pkg = require('../../package.json');
+
+type CommandConfig = {
```
matt-gadd: we probably want this in a separate file as it's the interface for consumers.

Tomdye:
@matt-gadd reopened with re-structuring

tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+	"version": "1.8.10",
```
agubler: `2.0.0`?

typings.json
```diff
@@ -0,0 +1,12 @@
+{
```
agubler: doesn't the template have a `name` property?

tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+	"compilerOptions": {
+		"declaration": false,
+		"module": "commonjs",
+		"noImplicitAny": true,
+		"outDir": "_build/",
+		"removeComments": false,
+		"sourceMap": true,
+		"target": "es6"
```
agubler: add `noImplicitThis` and `strictNullChecks`

src/logger.ts
```diff
@@ -0,0 +1,21 @@
+import { add, remove, transports, log } from 'winston';
+
+export interface VerboseOptions {
+	verbose?: boolean;
+}
+
+export default function setupLogger(verbose: boolean = false) {
+	remove(transports.Console);
```
agubler: How come you have to remove the `Console` transport first?

src/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+globby(globPaths(config)).then((paths) => {
+	verbose(`index - loading commands`);
+	const commandSet: command.CommandSet = new Set();
+	paths.forEach((path) => {
+		const commandConfig = command.load(path, commandSet);
+		const commandType = commandConfig.type;
+
+		if (commandsMap.has(commandType)) {
+			commandsMap.get(commandType).push(commandConfig);
```
matt-gadd: feels a bit side effecty to just do a `get` with a `push` rather than a `set` here? would this block not read nicer as:

```js
const commands = commandsMap.get(commandType) || [];
commandsMap.set(commandType, [...commands, commandConfig]);
```

src/command.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function getDescription(commandType: string, commandSubTypes: CommandConfig[]): string {
+	return commandSubTypes.length > 1 ?
+		`There are ${commandSubTypes.length} ${commandType} commands: ${commandSubTypes.map((commandSubType: CommandConfig) => commandSubType.name).join(', ')}` :
+		commandSubTypes[0].description;
+}
+
+export function register(commandsMap: CommandsMap): void {
+	for (let [ commandType, commandSubTypes ] of commandsMap.entries()) {
+		verbose(`index:registerCommand - commandType: ${commandType}, commandSubTypes: ${JSON.stringify(commandSubTypes)}`);
```
matt-gadd: have you thought about the structure of these logs? seems pretty verbose and the majority of it could be deduced rather than having these long manual templates in each function?

src/logger.ts
Tomdye: because it seems to be the only way to override it, actually, i think i can perhaps improve this now that `logger` provides the `log` functions as we can have our own logger rather than the global `winston` one

src/command.ts
```diff
@@ -0,0 +1,62 @@
+import { verbose } from './logger';
+import config from './config';
+import { command } from 'yargs';
+
+export interface Command {
```
matt-gadd: should this be in a separate file as it's an interface a consumer may use?

src/command.ts
```diff
@@ -0,0 +1,62 @@
...
+	name: string;
+	description: string;
+	register: Function;
```
matt-gadd: typings for these?

src/command.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export interface CommandConfig extends Command {
+	type: string;
+	subType: string;
+}
+
+export type CommandsMap = Map<string, CommandConfig[]>;
+
+export type CommandSet = Set<string>;
```
agubler: Do we need `CommandSet` and `CommandMap`?

src/command.ts
Tomdye: can't they use it from here?

src/command.ts
Tomdye: was going to look into that

src/command.ts
matt-gadd: i thought for external interfaces we preferred them in their own interfaces file?

agubler: We will actually create a `d.ts` file for all the modules with the module typings.

agubler: Misunderstood, it will still contain module interface as well as the public types/interfaces so doesn't change @matt-gadd's comment

matt-gadd: @agubler i get that but this is actually an external interface we want people to consume. we don't actually want them to consume this module. something like: https://github.com/dojo/routing/blob/master/src/interfaces.d.ts

src/logger.ts
agubler: Can you just do 

```js
  logger.configure({
    level: 'verbose',
    showLevel: false
  });
```

src/command.ts
```diff
@@ -0,0 +1,35 @@
...
+	};
+}
+
+export function getGroupDescription(group: string, commands: CommandsMap): string {
+	const commandNames: string[] = Array.from(commands.keys());
+	const numCommands = commandNames.length;
+	if (numCommands > 1) {
+		return `There are ${numCommands} ${group} commands: ${commandNames.join(', ')}`;
+	} else {
+		let { description } = <CommandWrapper> commands.get(commandNames[0]);
```
agubler: `let`?

tests/unit/all.ts
```diff
@@ -1 +1,3 @@
+import * as winston from 'winston';
```
matt-gadd: you can drop this now right?

src/interfaces.ts
```diff
@@ -0,0 +1,23 @@
+import { Argv, Yargs } from 'yargs';
+
+export interface RunResult {
+	success: boolean;
+	payload: any;
+}
+
+export interface TaskHelper {
```
matt-gadd: needs renaming. are we going to implement these in this PR?

src/CommandHelper.ts
```diff
@@ -0,0 +1,41 @@
...
+		this.groupsMap = groupsMap;
+	};
+	groupsMap: GroupsMap;
+	run(group: string, taskName?: string, args?: yargs.Argv): Promise<RunResult> {
+		const command = getCommand(this.groupsMap, group, taskName);
+		if (command) {
+			return command.run(new Helper(this, yargs, {}));
+		}
+		else {
+			return Promise.resolve({
```
matt-gadd: we should probably `reject` the Promise, we can probably drop the `success` from the `CommandResult` also.

src/command.ts
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+function pad(str: string, num: number): string {
+	const padding = new Array(num - str.length).join(' ');
+	return str + padding;
+}
+
+function getMultiCommandDescription(commands: CommandsMap): string {
+	const commandNames = Array.from(commands.keys(), (name) => name.length);
```
matt-gadd: `commandNameLengths`

src/CommandHelper.ts
```diff
@@ -0,0 +1,31 @@
...
+	const commandKey = commandName ? `${group}-${commandName}` : group;
+	return commandsMap.get(commandKey);
+}
+
+export default class implements CommandHelper {
+	constructor(commandsMap: CommandsMap, context: any) {
+		this.commandsMap = commandsMap;
+		this.context = context;
+	};
+	commandsMap: CommandsMap;
```
matt-gadd: should the `commandsMap` be on here? doesn't seem like a thing we should expose given the `exists` and `run` apis?

src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import config from './config';
+import loadCommands from './loadCommands';
+import registerCommands from './registerCommands';
+import { load, setSearchPrefix } from './command';
+const pkg = <any> require('../../package.json');
+
+async function init() {
+	updateNotifier(pkg, 0);
+	setSearchPrefix(config.searchPrefix);
+	const { commandsMap, yargsCommandNames } = await loadCommands(yargs, config, load);
```
agubler: why do you need to pass `load` to this function? Can it nor just import the `load` function?

src/loadCommands.ts
```diff
@@ -0,0 +1,46 @@
...
+		globby(globPaths, (<globby.Options> { ignore: '**/*.map' })).then((paths) => {
+			const commandsMap: CommandsMap = new Map();
+			const yargsCommandNames: YargsCommandNames = {};
+
+			paths.forEach((path) => {
+				const commandWrapper = load(path);
+				const { group, name } = commandWrapper;
+				const compositeKey = `${group}-${name}`;
+
+				// First of each type will be 'default' for now
```
matt-gadd: perhaps make this code a functions? then we at least have the concept of a `getDefaultCommand` to later refactor on?

src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+import updateNotifier from './updateNotifier';
+import config from './config';
+import loadCommands from './loadCommands';
+import registerCommands from './registerCommands';
+import { load, setSearchPrefix } from './command';
+const pkg = <any> require('../../package.json');
+
+async function init() {
+	updateNotifier(pkg, 0);
+	setSearchPrefix(config.searchPrefix);
```
agubler: For some reason I don't like this random function to set config as a global in another class - Would it be neater if the command class exported a function that you called to "init" and return the load function? 

so like (and you wouldn't need to do it in here)

```js
import initCommandLoader from './command';
const commandLoader = initCommentLoader(config);
```

Using `commandLoader` just like `load` now

src/interfaces.ts
```diff
@@ -0,0 +1,22 @@
+import { Argv, Yargs } from 'yargs';
+
+export interface RunResult {
```
agubler: Did we decided to leave the wrapper `RunResult` for the `payload` still?

tests/unit/command.ts
```diff
@@ -0,0 +1,61 @@
...
+			assert.equal(testName, commandWrapper.name);
+		},
+		'Should get description from loaded file'() {
+			assert.equal('test-description', commandWrapper.description);
+		},
+		'Should get register function from loaded file'() {
+			assert.isTrue(typeof commandWrapper.register === 'function');
+		},
+		'Should get run function from loaded file'() {
+			assert.isTrue(typeof commandWrapper.run === 'function');
```
matt-gadd: we should be able to check the equality of this against the real function.

agubler:
I generally think some more "doc" would be super useful

src/interfaces.ts
Tomdye: might need removing

src/index.ts
Tomdye: unable to stub it for testing otherwise

src/CommandHelper.ts
Tomdye: will mark as `private`

tests/unit/command.ts
Tomdye: cheers

src/index.ts
agubler: :trollface:

src/index.ts
Tomdye: sounds fair, so more of a factory

Tomdye:
@ant will add doc

agubler:
👍

PR #25, opened 2016-09-01
Move to using pkg-dir

Tomdye:


src/config.ts
```diff
@@ -1,4 +1,7 @@
-import { resolve } from 'path';
+import { join } from 'path';
+const pkgDir = require('pkg-dir');
```
matt-gadd: you're using `import` for deps everywhere else.

Issue #30, opened 2016-09-07
Allow the CLI to load esModules


Tomdye:
Commands written in TypeScript or Babel or similar add `__esModule` to the module to indicate that it came from an ESM module format. When this is present, load the `.default` command.

PR #31, opened 2016-09-07
Feature module format

Tomdye:
Fixes: https://github.com/dojo/cli/issues/30

- Checks for `__esModule` and loads `default` when present.
- Throws `Error` when bad modules loaded
- Catches errors from `load` function and consoles out

Tests Added

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -20,7 +20,8 @@ export function initCommandLoader(searchPrefix: string): (path: string) => Comma
 	const commandRegExp = new RegExp(`${searchPrefix}-(.*)-(.*)`);
 
 	return function load(path: string): CommandWrapper {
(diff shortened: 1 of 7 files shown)


src/command.ts
```diff
@@ -20,16 +20,27 @@ export function initCommandLoader(searchPrefix: string): (path: string) => Comma
...
-		return {
-			name,
-			group,
-			description,
-			register,
-			run
-		};
+		let module = require(path);
+
+		if (module.__esModule && module.default) {
```
matt-gadd: we should probably implement this in the loader at some point.

Tomdye:
issue raised in loader: https://github.com/dojo/loader/issues/87

Issue #34, opened 2016-09-08
`dojo -v` Should show version number of dojo-cli


Tomdye:

rishson:
@Tomdye isnt this covered in #11?

Tomdye:
Different issue @rishson. #11 refers to seeing the versions on installed dojo packages, this is showing the version of the `dojo-cli`

rishson:
OK - can you also get the `cli` to output the `cli`'s version when doing `dojo version` then so you get to see all the package versions and the cli version?

Tomdye:
yes, that will be fine

PR #35, opened 2016-09-08
added yargs version entry

Tomdye:
fixes: #34 

calling `dojo -v` will now output the version number of the cli. This is not the same as the upcoming `dojo version` command.


Tomdye:
This is good to go, it's complaining because an extra line is added to `index.ts` which is not tested.

PR #39, opened 2016-09-15
new: Updated readme with basic info, badges and default command usage.

rishson:


Tomdye:
it's `-help`. so `dojo command -help` not `dojo help command`

rishson:
Ohh - good spot.

readme.md
```diff
@@ -1,20 +1,68 @@
...
 
-TODO: Replace with a description of this package
+The CLI is the officially supported way to create and maintain Dojo 2 apps.
 
-## Features
+Its designed to save you time, by promoting a stadardised workflow, and automating away lots of the boring boilerplate tasks.
 
-TODO: Add sections on features of this package
+## Prerequisites
+You will need node V4 or v6+.
```
Tomdye: i thought we said node v6 for async await?

rishson: As mentioned yesterday, I'd like to update `engines` at the very least. Ideally, make `4+`, but understand (will put up with) 6+ until `TS 2.1`.

agubler: Even when TS2.1 comes out it means we will have to emit down to ES5 (currently emits down to ES6).... It comes back to, is it really unreasonable to ask developers to use a node v6+?

@rishson either way please for the sake of us all can you be consistent with your use of `v` and stick with either upper or lower case, not one of each. Kind Regards, my O.C.D.

Issue #37, opened 2016-09-13
Default / first group commands should run without the command name


Tomdye:
When registering `dojo-cli-example-command` it is expected that it should run when typing `dojo cli example`, but this is not currently the case.
Needs to be fixed.

PR #40, opened 2016-09-15
Fixed sub command issue

Tomdye:
Fixes: #37 

*Not written tests yet*

- Registers the default command so it's options show
- Runs the default command if only one command arg is present in `argv._`.

### DIFF
--- a/src/registerCommands.ts
+++ b/src/registerCommands.ts
@@ -53,8 +53,6 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 			// so we call default command, else, the subcommand will
 			// have been ran and we don't want to run the default.
 			if (argv._.length === 1) {
-				// Run the default command
-				console.log('running default command');
 				return defaultCommand.run(helper, argv);
 			}
 		});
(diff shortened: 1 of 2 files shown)


src/registerCommands.ts
```diff
@@ -42,6 +46,15 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 				);
 			});
 			return yargs;
+		},
+		(argv: Argv) => {
+			// argv._ is an array of commands.
```
agubler: What happens if I pass `dojo example -verbose` will that increase the length of the `argv`?

Tomdye: No, they become members of `argv`. ie.
calling `dojo example command -shout -verbose` creates an argv as follows

``` javascript
argv = {
   '_' : [ 'example', 'command' ],
   'shout' : true,
   'verbose' : true
}
```

Looks like that's how it's supposed to be used. Ensures that `options` such as `verbose` don't overwrite command names.

agubler: Okay great thanks.

Tomdye: good call asking though, it made me realise it's be design

Issue #52, opened 2016-09-30
Does not load globally installed commands


agubler:
if a user has `dojo-cli` and for example `dojo-cli-create-app` both installed locally the `cli` should load the `create-app` command but it doesn't.

PR #53, opened 2016-09-30
Feature find global modules

Tomdye:
Fixes: https://github.com/dojo/cli/issues/52
Needs to loon one dir up because the `dojo.js` bin file is in a `bin` dir.


Tomdye:
Needs to have a test added in `config.ts`, will need to abstract away `__dirname` which I can now do via `mockery.js`.

agubler:
let's have a test :)

Tomdye:
@agubler tided up kitchen sink

PR #55, opened 2016-09-30
Increase code coverage for CLI

Tomdye:
Using the `dirname` import has made it possible to add tests for `index.ts` and `updateNotifier.ts`


Tomdye:
@agubler comments addressed

PR #64, opened 2016-10-12
Adding built-in version command, fixes #11

### LINKED ISSUE #11: version command
This should output the versions of all registered groups, e.g.

```
dojo version

The currently installed groups are:

build (dojo/cli-build-webpack) 0.23.121
build (someAuthor/dojo-cli-MyBuildCommand) 1.2.3
test (dojo/cli-test) 0.3.22
...

You are currently running dojo-cli 2.1.223
```

Users can then copy and paste this info into issues raised on GH. This allows a user to provide version information as easily as possible, thus increasing the likelihood that they will include this information in GH issues.

rorticus:
Adding `version` command. This command is built into cli core and will show the version of all the registered commands.

Usage:

```
$ dojo version
```

Example output with no commands,

```
There are no registered commands available.

You are currently running dojo-cli 2.0.0-pre
```

Example output with commands,

```
The currently installed groups are:

create (dojo-cli-create-app) 2.0.0-alpha.2

You are currently running dojo-cli 2.0.0-pre
```

What the help looks like,

```
dojo help

Usage: dojo <group> <command> [options]

Hey there, here are all the things you can do with dojo-cli:

Commands:
  create   Scaffolds a new command
  version  List versions of registered commands

Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]

For more information on any of these commands just run them with '-h'.

e.g. 'dojo run -h' will give you the help for the 'run' group of commands.
```

### DIFF
--- a/src/registerCommands.ts
+++ b/src/registerCommands.ts
@@ -2,7 +2,8 @@ import { Yargs, Argv } from 'yargs';
 import { getGroupDescription, CommandsMap, CommandWrapper } from './command';
 import CommandHelper from './CommandHelper';
 import Helper from './Helper';
-import { helpUsage, helpEpilog } from './text';
+import { helpUsage, helpEpilog, versionDescription } from './text';
+import createVersionsString from './version';
 import * as chalk from 'chalk';
 
 export interface YargsCommandNames {
@@ -69,6 +70,8 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 		.help('h')
 		.alias('h', 'help')
 		.alias('v', 'version')
+		.version(() => createVersionsString(commandsMap))
+		.command('version', versionDescription, (yarts) => yargs, () => console.log(createVersionsString(commandsMap)))
 		.strict()
 		.argv;
 }
--- a/src/text.ts
+++ b/src/text.ts
@@ -1,4 +1,4 @@
-import { bold } from 'chalk';
+import { bold, yellow } from 'chalk';
 
 export const helpUsage = `${bold('dojo help')}
 
@@ -9,3 +9,18 @@ Hey there, here are all the things you can do with dojo-cli:`;
 export const helpEpilog = `For more information on any of these commands just run them with '-h'.
 
 e.g. 'dojo run -h' will give you the help for the 'run' group of commands.`;
+
+export const versionCurrentVersion = `
+You are currently running dojo-cli {version}
+`;
+
+export const versionDescription = 'List versions of registered commands';
+
+export const versionNoRegisteredCommands = `
+There are no registered commands available.`;
+
+export const versionNoVersion = yellow('package.json missing');
+
+export const versionRegisteredCommands = `
+The currently installed groups are:
+`;
--- a/src/version.ts
+++ b/src/version.ts
@@ -11,17 +11,29 @@ import { join } from 'path';
 const pkgDir = require('pkg-dir');
 const packagePath = pkgDir.sync(dirname);
 
+/**
+ * The details of one command group's module.
+ */
 export interface ModuleVersion {
 	name: string;
 	version: string;
 	group: string;
 }
 
+/**
+ * Important information to be retrieved from a module's package.json
+ */
 export interface PackageDetails {
 	name: string;
 	version: string;
 }
 
+/**
+ * Read information about a package/module, if available, or return default values.
+ *
+ * @param {string} packageDir The directory containing the package.json file.
+ * @returns {{name: (any|string), version: any}}
+ */
 export function readPackageDetails(packageDir: string): PackageDetails {
 	let data: any = {};
 	try {
@@ -36,7 +48,14 @@ export function readPackageDetails(packageDir: string): PackageDetails {
 	};
 }
 
-export function buildVersions(commandsMap: CommandsMap) {
+/**
+ * Iterate through a CommandsMap and retrieve the module details of each module referenced in the
+ * CommandsMap. The returned list is sorted in alphabetical order, by group.
+ *
+ * @param {CommandsMap} commandsMap
+ * @returns {{name, version, group}[]}
+ */
+export function buildVersions(commandsMap: CommandsMap): ModuleVersion[] {
 	/*
 	 * commandsMap comes in as a map of [command-name, command]. The command has a default command,
 	 * the map will actually contain two entries for one command, on for the default command, one for the real,
@@ -59,6 +78,13 @@ export function buildVersions(commandsMap: CommandsMap) {
 		}).sort((a, b) => a.group > b.group ? 1 : -1);
 }
 
+/**
+ * Returns a string describing the command group, module name, and module version of each
+ * command referenced in a specified CommandsMap. This is used to print the the help string.
+ *
+ * @param {CommandsMap} commandsMap
+ * @returns {string}
+ */
 export default function createVersionsString(commandsMap: CommandsMap): string {
 	const myPackageDetails = readPackageDetails(packagePath),
 		commands = buildVersions(commandsMap);
--- a/src/version.ts
+++ b/src/version.ts
@@ -1,0 +1,80 @@
+import { CommandsMap } from './command';
+import dirname from './dirname';
+import {
+	versionNoRegisteredCommands,
+	versionRegisteredCommands,
+	versionCurrentVersion,
+	versionNoVersion
+} from './text';
+import { join } from 'path';
+
+const pkgDir = require('pkg-dir');
+const packagePath = pkgDir.sync(dirname);
+
+export interface ModuleVersion {
+	name: string;
+	version: string;
+	group: string;
+}
+
+export interface PackageDetails {
+	name: string;
+	version: string;
+}
+
+export function readPackageDetails(packageDir: string): PackageDetails {
+	let data: any = {};
+	try {
+		data = require(join(packageDir, 'package.json'));
+	} catch (e) {
+		data = {};
+	}
+
+	return {
+		name: data.name || packageDir,
+		version: data.version || versionNoVersion
+	};
+}
+
+export function buildVersions(commandsMap: CommandsMap) {
+	/*
+	 * commandsMap comes in as a map of [command-name, command]. The command has a default command,
+	 * the map will actually contain two entries for one command, on for the default command, one for the real,
+	 * expanded, command.
+	 *
+	 * Loop over commandsMap and create a new map with one entry per command, then loop over each entry and extract
+	 * the package details.
+	 */
+	const consolidatedCommands = [ ...new Map<string, string>([ ...commandsMap ]
+		.map(([, command]) => <[string, string]> [ command.path, command.group ])) ];
+
+	return consolidatedCommands.map(([path, group]) => {
+			const { name, version } = readPackageDetails(path);
+
+			return {
+				name,
+				version,
+				group
+			};
+		}).sort((a, b) => a.group > b.group ? 1 : -1);
+}
+
+export default function createVersionsString(commandsMap: CommandsMap): string {
+	const myPackageDetails = readPackageDetails(packagePath),
+		commands = buildVersions(commandsMap);
+	let output = '';
+
+	if (commands.length) {
+		output += versionRegisteredCommands;
+		output += '\n'
+			+ commands.map((command) => `${command.group} (${command.name}) ${command.version}`).join('\n')
+			+ '\n';
+	}
+	else {
+		output += versionNoRegisteredCommands;
+	}
+
+	output += versionCurrentVersion.replace('\{version\}', myPackageDetails.version);
+
+	return output;
+}
--- a/tests/unit/text.ts
+++ b/tests/unit/text.ts
@@ -4,12 +4,23 @@ const text = require('intern/dojo/node!../../src/text');
 
 registerSuite({
 	name: 'text',
-	'Should export helpUsage'() {
-		assert.isNotNull(text.helpUsage);
-		assert.equal('string', typeof text.helpUsage);
-	},
-	'Should export helpEpilog'() {
-		assert.isNotNull(text.helpEpilog);
-		assert.equal('string', typeof text.helpEpilog);
-	}
+	exports: (function (exportedStrings) {
+		const tests: (() => void)[] = [];
+
+		exportedStrings.forEach(function (exportedString) {
+			tests.push(function () {
+				assert.isNotNull(text[exportedString]);
+				assert.isString(text[exportedString]);
+			});
+		});
+
+		return tests;
+	})([
+		'helpUsage',
+		'helpEpilog',
+		'versionDescription',
+		'versionNoRegisteredCommands',
+		'versionRegisteredCommands',
+		'versionCurrentVersion'
+	])
 });
--- a/tests/unit/version.ts
+++ b/tests/unit/version.ts
@@ -1,0 +1,88 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+import { CommandsMap, CommandWrapper } from '../../src/command';
+import { getCommandWrapperWithConfiguration } from '../support/testHelper';
+import { versionNoRegisteredCommands, versionNoVersion, versionRegisteredCommands } from '../../src/text';
+const { readPackageDetails, buildVersions, 'default': createVersionsString } = require('intern/dojo/node!../../src/version');
+
+registerSuite({
+	name: 'text',
+
+	readPackageDetails: {
+		validPackage() {
+			const details = readPackageDetails('../tests/support/valid-package');
+
+			assert.equal(details.name, 'Test Package 1');
+			assert.equal(details.version, '1.0.0');
+		},
+
+		missingPackage() {
+			const details = readPackageDetails('../tests/support');
+
+			assert.equal(details.name, '../tests/support');
+			assert.equal(details.version, versionNoVersion);
+		}
+	},
+
+	buildVersions() {
+
+		const commandWrapper1 = getCommandWrapperWithConfiguration({
+				group: 'apple',
+				name: 'test',
+				path: '../tests/support/valid-package'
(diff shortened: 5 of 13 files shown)


src/text.ts
```diff
@@ -9,3 +9,16 @@ Hey there, here are all the things you can do with dojo-cli:`;
 export const helpEpilog = `For more information on any of these commands just run them with '-h'.
 
 e.g. 'dojo run -h' will give you the help for the 'run' group of commands.`;
+
+export const versionCurrentVersion = `
```
agubler: Just a question do we prefer this formatting of strings over using `\n` inline?

src/version.ts
```diff
@@ -0,0 +1,71 @@
...
+		name: data.name || packageDir,
+		version: data.version || 'N/A'
+	};
+}
+
+export function buildVersions(commandsMap: CommandsMap) {
+	const commands: ModuleVersion[] = [];
+	const moduleGroups = new Map<string, string>();
+
+	commandsMap.forEach(function (command) {
```
agubler: I think we are trying to be consistent with inline anonymous function to use arrows (@kitsonk said he would update the style guide).

src/version.ts
```diff
@@ -0,0 +1,71 @@
...
+	group: string;
+}
+
+interface PackageDetails {
+	name: string;
+	version: string;
+}
+
+export function readPackageDetails(packageDir: string): PackageDetails {
+	const data = require(join(packageDir, 'package.json'));
```
agubler: What happens if there is no `package.json`?

src/version.ts
```diff
@@ -0,0 +1,71 @@
...
+
+	return commands;
+}
+
+export default function createVersionCommand(commandsMap: CommandsMap) {
+	return function () {
+		const myPackageDetails = readPackageDetails(packagePath);
+		const commands = buildVersions(commandsMap);
+
+		if (commands.length === 0) {
```
agubler: `if (commands.length) {`?

src/registerCommands.ts
```diff
@@ -69,6 +70,9 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 		.help('h')
 		.alias('h', 'help')
 		.alias('v', 'version')
```
agubler: is there a way to remove `-v` and `--version` now that we have the `dojo version` command?

src/version.ts
```diff
@@ -0,0 +1,71 @@
...
+export function readPackageDetails(packageDir: string): PackageDetails {
+	const data = require(join(packageDir, 'package.json'));
+
+	return {
+		name: data.name || packageDir,
+		version: data.version || 'N/A'
+	};
+}
+
+export function buildVersions(commandsMap: CommandsMap) {
```
matt-gadd: is the intermediary map needed in this function? not tried but would something like the following be equivalent?
```ts
export function buildVersions(commandsMap: CommandsMap): ModuleVersion[] {
	return [ ...commandsMap ].map(([ , command ]) => {
		const { name, version } = readPackageDetails(command.path);
		return { 
				name, 
				version, 
				group: command.group 
		};
	}).sort((a, b) => a.group > b.group ? 1 : -1);
}
```

tests/unit/version.ts
```diff
@@ -0,0 +1,96 @@
...
+		afterEach: function () {
+			consoleStub.restore();
+		},
+
+		'no commands'() {
+			let commandMap: CommandsMap = new Map<string, CommandWrapper>();
+
+			createVersionCommand(commandMap)();
+
+			assert.isTrue(loggedOutput.indexOf('no registered') >= 0);
```
matt-gadd: I think this might be a little cleaner if you just used `calledWith`, and imported the text from the text module to assert, what ya think?
```ts
import { versionNoRegisteredCommands } from '../../src/text';
```
```
assert.isTrue(consoleStub.calledWith(versionNoRegisteredCommands);
```

tests/unit/version.ts
```diff
@@ -0,0 +1,96 @@
...
+		assert.equal(commands[1].version, 'N/A');
+		assert.equal(commands[1].group, 'test2');
+	},
+
+	'createCommand': {
+		beforeEach: function () {
+			loggedOutput = '';
+
+			consoleStub = stub(console, 'log', function () {
+				for (let i = 0; i < arguments.length; i++) {
```
matt-gadd: if you use rest params rather than `arguments` you can type the args as a string rather than having to explicitly use `String` later. You could also just `join` the args then you can drop the loop.

src/version.ts
```diff
@@ -0,0 +1,71 @@
...
+export default function createVersionCommand(commandsMap: CommandsMap) {
+	return function () {
+		const myPackageDetails = readPackageDetails(packagePath);
+		const commands = buildVersions(commandsMap);
+
+		if (commands.length === 0) {
+			console.log(versionNoRegisteredCommands);
+		}
+		else {
+			console.log(versionRegisteredCommands);
```
matt-gadd: we should potentially look at wrapping the `console.log` in a function in a separate module at some point. that way we can more safely stub it and not lose console.log functionality in our tests.

src/registerCommands.ts
rishson: we should keep the `-v` alias of `version`

agubler: but that just prints out the cli version and would confuse users imo. They be like why do I haz multiple ways of getting versions

src/text.ts
rorticus: Yeah.. I could go either way on this one. I actually originally had `\n` in there and then switched it over to template strings to be consistent with the other strings in the file.  Just let me know which to use and I'll update my brain accordingly :)

src/version.ts
agubler: ah ha okay, I think the latest guidance is described on this ticket https://github.com/dojo/meta/issues/45

src/version.ts
rorticus: I've also seen other modules take a "logging function" as a parameter.

tests/unit/version.ts
rorticus: 👍

tests/unit/version.ts
rorticus: I like that! I didn't know you could do that. hehe

src/registerCommands.ts
rishson: `-v` and `-h` are just common aliases, don't see why we wouldn't have them just redirect to `versions` and `help` respectively.

rishson: Contrast to something like `npm` where `-v` gives you the version of `npm`, whereas `version` gives you the versions of all deps in the current proj.

agubler: Perhaps we are crossing wires here..... So we have implemented a command `version` that returns information about the `CLI` version and the version of all the registered commands, which is called by `dojo version`.

Also, as well as this there the option `-v` or `--version` that provides just the version of the `CLI` itself (pre existing before this PR), called by `dojo -v` or `dojo --version`.

To me as a user it would be weird to have to `version` command/options especially if they output different information...

rishson: Don't think were xing wires. The current behaviour is as per `npm` (`npm -v && npm version`). A `-v` option is standard so we should have one. A `version` command is really useful, so we should have one. Your point about the 2 things being the same/different, I don't really mind. We can stick to a convention of sorts and be `npm` like, or we can have them consistent and both output versions of the commands and the cli. Consistency is def worthwhile, so why don't we just redirect `-v` to versions?

agubler: Is that possible? If so I don't mind that, but is should have the same output as the `version` command imo.

src/version.ts
rorticus: OK so I'm adding a check for `package.json`.. I plan on just using an empty object if the file isn't there so that it'll just use the defaults. Do you think I should `console.warn` about this, or just silently continue?

rishson: it should warn (or similar). At the mo, we only allow `npm` modules as commands, so something if def wrong if no `package.json`.

rorticus: What if instead of `N/A` I default to a warning message, i.e.,

```
create (node_modules/dojo-cli-create-app) package.json not found
```

Where `package.json not found` is maybe in yellow?

src/registerCommands.ts
```diff
@@ -2,7 +2,8 @@ import { Yargs, Argv } from 'yargs';
 import { getGroupDescription, CommandsMap, CommandWrapper } from './command';
 import CommandHelper from './CommandHelper';
 import Helper from './Helper';
-import { helpUsage, helpEpilog } from './text';
+import {helpUsage, helpEpilog, versionDescription} from './text';
```
rorticus: Updated!

src/registerCommands.ts
rorticus: OK I was able to consolidate everything into using `-v`, `--version` and the `version` command.

src/registerCommands.ts
```diff
@@ -69,6 +70,9 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 		.help('h')
 		.alias('h', 'help')
 		.alias('v', 'version')
+		.command('version', versionDescription, function (yargs: Yargs) {
```
rorticus: All the arrows.

src/version.ts
rorticus: Added support for missing `package.json`

src/version.ts
rorticus: Updated

src/version.ts
rorticus: I had to add another map in here ( the command map contains duplicates ), but otherwise this was good to go :)

src/version.ts
rorticus: This was sort of solved because I needed to refactor the code to return a string anyways..

src/version.ts
matt-gadd: duplicates of what? and if so should we not fix the command map, as it sounds incorrect there if thats the case?

rorticus: Because I'm looping over the commands (and don't care about the keys), I'm getting an extra copy of commands that have a default. (i.e., `dojo-cli-create-app` has two commands, `create`, and `create-app`).

Here is a copy of the command map with just `dojo-cli-create-app` installed,

```
[ [ 'create',
    { name: 'app',
      group: 'create',
      description: 'Scaffolds a new command',
      register: [Function: default_1],
      run: [Function: default_1],
      path: '/Users/rorymulligan/Desktop/dojo2/cli/node_modules/dojo-cli-create-app' } ],
  [ 'create-app',
    { name: 'app',
      group: 'create',
      description: 'Scaffolds a new command',
      register: [Function: default_1],
      run: [Function: default_1],
      path: '/Users/rorymulligan/Desktop/dojo2/cli/node_modules/dojo-cli-create-app' } ] ]
```

tests/unit/text.ts
```diff
@@ -4,12 +4,23 @@ const text = require('intern/dojo/node!../../src/text');
...
-		assert.isNotNull(text.helpEpilog);
-		assert.equal('string', typeof text.helpEpilog);
-	}
+	exports: (function (exportedStrings) {
+		const tests: (() => void)[] = [];
+
+		exportedStrings.forEach(function (exportedString) {
+			tests.push(function () {
+				assert.isNotNull(text[exportedString]);
+				assert.equal('string', typeof text[exportedString]);
```
agubler: what about `assert.isString(text[exportedString]);`?

src/version.ts
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return {
+		name: data.name || packageDir,
+		version: data.version || versionNoVersion
+	};
+}
+
+export function buildVersions(commandsMap: CommandsMap) {
+	return [ ...new Map<string, string>([ ...commandsMap ].map(([, command]) => <[string, string]> [ command.path, command.group ])) ]
```
agubler: This would probably be easier to understand if it was broken up a bit...? Perhaps a comment also describing the reason for the `Map`.

agubler:
Can we add some general JSDoc, especially for exported `interfaces` / `function` / `types`?

src/version.ts
matt-gadd: Wow thats pretty annoying that they are just registered under keys like that and not 2 separate maps or a key that at least signifies its a default. Given it seems like an important differentiation we'd likely need again, how about adding an `isDefault` to the CommandWrapper? That way this becomes a fair bit cleaner with a `filter`.

src/registerCommands.ts
```diff
@@ -2,13 +2,26 @@ import { Yargs, Argv } from 'yargs';
...
-import { helpUsage, helpEpilog } from './text';
+import { helpUsage, helpEpilog, versionDescription } from './text';
+import createVersionString from './version';
 import * as chalk from 'chalk';
 
 export interface YargsCommandNames {
 	[property: string]: string[];
 };
 
+export function createVersionStringFn(commandsMap: CommandsMap) {
```
matt-gadd: do we need this? it seems like an extra level of indirection?

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+import { CommandsMap, CommandWrapper } from '../../src/command';
+import { getCommandWrapperWithConfiguration } from '../support/testHelper';
+import { versionNoRegisteredCommands, versionNoVersion, versionRegisteredCommands } from '../../src/text';
+const {readPackageDetails, buildVersions, 'default': createVersionString} = require('intern/dojo/node!../../src/version');
```
agubler: Why the `require('intern/dojo/node!...`?

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
...
+
+			let commandMap: CommandsMap = new Map<string, CommandWrapper>([
+				['test2', commandWrapper2],
+				['test', commandWrapper1]
+			]);
+
+			let output = createVersionString(commandMap);
+
+			assert.isTrue(output.indexOf(versionRegisteredCommands) >= 0);
+			assert.isTrue(output.indexOf('test') >= 0);
```
matt-gadd: i'd feel a lot more confident if these were more exact assertions. at the moment if the `versionRegisteredCommands` static text contained the word `test` this assertion would pass.

tests/unit/version.ts
agubler: Okay I see it needs the node require to run, but can we clean up the styling for the `{`/`}` and also be consistent on the name `createVersionString` as it is exported as `createVersionsString` (extra `s`) so I couldn't find it when searching the PR.

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
...
+				});
+
+			let commandMap: CommandsMap = new Map<string, CommandWrapper>([
+				['test2', commandWrapper2],
+				['test', commandWrapper1]
+			]);
+
+			let output = createVersionString(commandMap);
+
+			assert.isTrue(output.indexOf(versionRegisteredCommands) >= 0);
```
agubler: Ignore this, given they are text assertions (as per @matt-gadd's comment below), we should be using a more exact assertion as suggested.

src/registerCommands.ts
rorticus: OK, these methods only existed to make the logic testable. I was able to get rid of these by using the yargs stub stuff.

tests/unit/text.ts
rorticus: Didn't know `isString` existed, it's perfect!

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
...
+import { getCommandWrapperWithConfiguration } from '../support/testHelper';
+import { versionNoRegisteredCommands, versionNoVersion, versionRegisteredCommands } from '../../src/text';
+const {readPackageDetails, buildVersions, 'default': createVersionString} = require('intern/dojo/node!../../src/version');
+
+registerSuite({
+	name: 'text',
+
+	readPackageDetails: {
+		validPackage() {
+			let details = readPackageDetails('../tests/support/valid-package');
```
rorticus: My bad :)

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
...
+
+		missingPackage() {
+			let details = readPackageDetails('../tests/support');
+
+			assert.equal(details.name, '../tests/support');
+			assert.equal(details.version, versionNoVersion);
+		}
+	},
+
+	buildVersions() {
```
rorticus: I rearranged the examples being passed in, as well as gave them more easily recognizable alphabetical names, so now the sorting is performed and is verified later down (not explicitly, but implicitly by the `[0]` and `[1]` indexes)

tests/unit/version.ts
```diff
@@ -0,0 +1,88 @@
...
+		assert.equal(commands[1].name, '../tests/support');
+		assert.equal(commands[1].version, versionNoVersion);
+		assert.equal(commands[1].group, 'test2');
+	},
+
+	'createCommand': {
+		'no commands'() {
+			let commandMap: CommandsMap = new Map<string, CommandWrapper>();
+
+			assert.isTrue(createVersionString(commandMap).indexOf(versionNoRegisteredCommands) >= 0);
```
rorticus: I didn't know `include` was a thing. I'm learning more about these available assertions every day! :)

tests/unit/version.ts
rorticus: Increased exactedness!

src/text.ts
rishson: We don't have a standard on this yet, I think. For now, I think that multiline template strings are fine and IMO prefered.

src/registerCommands.ts
rishson: 👍

src/text.ts
rishson: 👍

rorticus:
@agubler says,

> Can we add some general JSDoc, especially for exported interfaces / function / types?

Done!

Issue #73, opened 2016-11-09
Refactor to separate out concerns of 'tty processing', 'loading commands' and 'running commands'


rishson:
I think it would improve testability if we separated the code into:

1) tty processing using `yargs`
2) installed commands resolution and loading
3) running commands (in-built and installed)

Commands should not really be aware of their invocation context, i.e. that they are being run in a terminal.
This would make commands easier to test.
At the moment we have:
a command loader implementation (`command`) that knows about screen width.
a command loader (`loadCommands`) that knows about `yargs`
a command runner (`registerCommands`) that gets passed a command loader implementation for non-loader functionality.

I propose the following refactor:

`command` -> `commandLoader`
`loadCommands` -> `runCommandLoader`
`registerCommands` -> `runCommands`

where

`index.ts` is the only `tty` aware module
`commandLoader` is the only file system aware module
`runCommandLoader` becomes a simple loader executor
`runCommands` becomes a simple command executor
in-built commands become async commands in a `/commands` subdir and simply get added to the list of commands to run.
command specific text is kept with the command, rather than collected up into a cross-command `text.ts` file.
split `commnds/version` into the `Command` and the outdated check.

In addition, unit tests contain a test for `index.ts`. This is actually a functional test and should be moved into a function test dir, and should be tested against without mocks.

I think this is a cleaner separation of concerns that should be easier to test and extend.

kitsonk:
References #140

PR #75, opened 2016-11-17
Implement version --outdated.

### LINKED ISSUE #67: Hide built-in groups such as 'versions` from being listed in the installed commands.
We should hide built in groups such as `versions` so that we see:

```
$ dojo
dojo help

Usage: dojo <group> <command> [options]

Hey there, here are all the things you can do with dojo-cli:

Commands:
  build    create a build of your application

Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]

For more information on any of these commands just run them with '-h'.

e.g. 'dojo run -h' will give you the help for the 'run' group of commands.
```

instead of:

```
$ dojo
dojo help

Usage: dojo <group> <command> [options]

Hey there, here are all the things you can do with dojo-cli:

Commands:
  build    create a build of your application
  version  List versions of registered commands

Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]

For more information on any of these commands just run them with '-h'.

e.g. 'dojo run -h' will give you the help for the 'run' group of commands.
```

### LINKED ISSUE #68: version --outdated command
Refs #49 .
Implement `version --outdated`

rishson:
Fixes #67, #68, addresses some of #73.
Adding this behaviour has meant moving the existing `version` functionality to a `Command` because the `--outdated` check is async.
This has caused a limited refactor.
Where possible, I have not altered the code layout or functionality to keep this PR small, however, I have done the following:

- fixed all `tslint` errors.
- increased the coverage as far as I can

Getting code coverage from 99.87% to 100% will require introducing extra dependencies or re-writing how the `cli` main entry func works, so I have not done this.

The functionality is limited to using `npm` as the `cli` module repository.
 The `command` interface now has `group` and `name` as optional properties, because these are specified by the `version` in-built command, rather than being derived from the command's enclosing directory name.


src/CommandHelper.ts
```diff
@@ -12,7 +12,7 @@ function getCommand(commandsMap: CommandsMap, group: string, commandName?: strin
  * CommandHelper class which is passed into each command's run function
  * allowing commands to call one another. Provides 'run' and 'exists' functions
  */
-export default class implements CommandHelper {
+export default class SomeClass implements CommandHelper {
```
agubler: `SomeClass`?

rishson: Yeah - wasn't sure what to put here. This was basically for WebStorm as it cant process default nameless classes, so I chose a name that showed the name is optional. Also thought of using `clazz` etc..

Tomdye: Call it `CLICommandHelper` then

src/command.ts
```diff
@@ -23,12 +22,10 @@ export function initCommandLoader(searchPrefix: string): (path: string) => Comma
...
 		let module = require(path);
 
-		if (module.__esModule && module.default) {
-			module = module.default;
-		}
-
-		if (module.description && module.register && module.run) {
-			const { description, register, run } = <Command> module;
+		try {
+			const command = convertModuleToCommand(module);
```
Tomdye: doesn't look like `command` is being used anywhere so can just do
```
const {description, register, run} = convertModuleToCommand(module);
```

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+export const versionNoVersion = yellow('package.json missing');
+export const versionRegisteredCommands = `
+The currently installed groups are:
+`;
+const INBUILT_COMMAND_VERSION = '__IN_BUILT_COMMAND__';
+
+/**
+ * The details of one command group's module.
+ */
+export interface ModuleVersion {
```
Tomdye: feels like this should be elsewhere such that it can be used be all built in commands

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
+import {CommandsMap } from '../command';
```
Tomdye: I think this file would be better split into smaller files under a `version` directory, it's pretty big at the moment, also you're using functions / types before they're declared and the ordering / placement of your types and exports is wrong.

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+function buildVersions(commandsMap: CommandsMap): Promise<any> {
+	/*
+	 * commandsMap comes in as a map of [command-name, command]. The command has a default command,
+	 * the map will actually contain two entries for one command, on for the default command, one for the real,
+	 * expanded, command.
+	 *
+	 * Loop over commandsMap and create a new map with one entry per command, then loop over each entry and extract
+	 * the package details.
+	 */
+	const consolidatedCommands = [ ...new Map<string, string>([ ...commandsMap ]
```
Tomdye: Not very clear what's going on here, when converting an iterator type to an array and performing a `map` I find it's better to use `Array.from(iterator, mapFunction)`

src/index.ts
```diff
@@ -1,10 +1,11 @@
 import * as yargs from 'yargs';
 import updateNotifier from './updateNotifier';
 import config from './config';
-import loadCommands from './loadCommands';
+import {loadCommands } from './loadCommands';
```
Tomdye: spacing

src/index.ts
```diff
@@ -1,10 +1,11 @@
...
 import registerCommands from './registerCommands';
-import { initCommandLoader } from './command';
+import { initCommandLoader, createBuiltInCommandLoader } from './command';
 import { join } from 'path';
+import { enumerateInstalledCommands, enumerateBuiltInCommands} from './loadCommands';
```
Tomdye: second import from `loadCommands`, should combine them

src/Helper.ts
```diff
@@ -1,19 +1,22 @@
 import { CommandHelper, Helper } from './interfaces';
 import { Yargs } from 'yargs';
+import {CommandsMap} from './command';
```
Tomdye: spacing

tests/unit/commands/version.ts
```diff
@@ -0,0 +1,166 @@
...
+
+		const commandMap: CommandsMap = new Map<string, CommandWrapper>([
+			['installedCommand1', installedCommandWrapper]
+		]);
+
+		const helper = {commandsMap: commandMap, command: 'version' };
+		return moduleUnderTest.run(helper, {'outdated': true}).then(() => {
+			assert.equal('Fetching latest version information...', (<sinon.SinonStub> console.log).args[0][0]);
+			assert.equal(expectedOutput, (<sinon.SinonStub> console.log).args[1][0]);
+		}, () => {
```
Tomdye: be aware that our standards dictate that arrow functions should only be used when we're using `this` within the function. Test a fail via `assert.fail(null, null, 'Reason for failure')'` - that is how it is done elsewhere

Tomdye:
Initial comments raised, still needs a more thorough review.

src/CommandHelper.ts
rishson: Yep - will do

src/command.ts
rishson: As discussed, I think a useful coding standard is to have redundant local variables when doing destructuring. The reason I like this is that if you reference a property in your destructuring that does not exist on the returned object, then the error message is pretty obtuse. A redundant local will be inlined by the compiler anyway and its much easier to debug.

src/commands/version.ts
rishson: Potaytoe/potartoe - I really like the rest/spread syntax and like the `Map = Map + Map` ness.

src/index.ts
rishson: ? whats wrong with importing 2 funcs from a module?

tests/unit/commands/version.ts
rishson: OK - weird standard, but whatevs.

src/commands/version.ts
rishson: As discussed - only used here.

src/index.ts
rishson: 👍

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+	helper.yargs.option('outdated', {
+		alias: 'outdated',
+		describe: 'Output a list of installed commands that can be updated to a more recent stable version.',
+		demand: false,
+		type: 'string'
+	});
+	return helper.yargs;
+}
+
+export interface VersionArgs extends Argv {
```
agubler: again why exported?

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+ * Is the command a built in command as opposed to an installed command
+ * @param commandPath path to the command module
+ * @returns {boolean}
+ */
+function isBuiltInCommand(commandPath: string): boolean {
+	/*__dirname seems best as the only way to truly know if a command is built in, is by location.
+	* Since this module is a built in command, we can use our location.
+	* This was preferable to using packageDir and relative paths, because we may alter where we build to (_build/src...)
+	*/
+	return commandPath.startsWith(join(__dirname));
```
agubler: why `join`?

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+ * @param {string} packageDir The directory containing the package.json file.
+ * @returns {{name: (any|string), version: any}}
+ */
+function readPackageDetails(packageDir: string): PackageDetails {
+	let data: any = {};
+	try {
+		data = require(join(packageDir, 'package.json'));
+	} catch (e) {
+		// rather than add another prop to Command, declare the command to be builtin by setting its version
+		if (isBuiltInCommand(packageDir)) {
```
agubler: Should we check this before we try and require the package.json and catch an error?

src/commands/version.ts
rishson: Can you be specific on the mis-ordering and 'use-before' so that I can fix please?

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	// create fake manifest (package.json) with just the dev-dependencies that we want to check
+	const manifest = {
+		devDependencies: deps
+	};
+
+	return new Promise((resolve, reject) => {
+		// we want to fetch the latest stable version for our devDependencies
+		david.getUpdatedDependencies(manifest, { dev: true, stable: true }, function (er: any, deps: {[dependencyName: string]: {stable: string }}) {
+			if (er) {
```
agubler: `er`, we can be perhaps a little better :)

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+		});
+
+	// create fake manifest (package.json) with just the dev-dependencies that we want to check
+	const manifest = {
+		devDependencies: deps
+	};
+
+	return new Promise((resolve, reject) => {
+		// we want to fetch the latest stable version for our devDependencies
+		david.getUpdatedDependencies(manifest, { dev: true, stable: true }, function (er: any, deps: {[dependencyName: string]: {stable: string }}) {
```
agubler: lets create a type/interface for the `deps`

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+ * Returns a string describing the command group, module name, and module version of each
+ * command referenced in a specified CommandsMap. This is used to print the string.
+ *
+ * @param {CommandsMap} commandsMap maps of commands to output the versions for
+ * @param {boolean} checkOutdated should we check if there is a later stable version available for the command
+ * @returns {string} the stdout output
+ */
+function createVersionsString(commandsMap: CommandsMap, checkOutdated: boolean): Promise<string> {
+	const packagePath = pkgDir.sync(__dirname);
+
```
agubler: more random line breaks 😱

src/commands/version.ts
agubler: Then do not export it?

rishson: Comment was on a private magic value, not the interface, I believe.

src/command.ts
```diff
@@ -39,13 +36,52 @@ export function initCommandLoader(searchPrefix: string): (path: string) => Comma
 				run,
 				path
 			};
+		} catch (err) {
+			throw new Error(`Path: ${path} returned module that does not satisfy the Command interface`);
```
rishson: yup

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
+import {CommandsMap } from '../command';
+import { Command, Helper } from '../interfaces';
+
```
rishson: Was splitting imports by:

- own code
- env specific, e.g. node modules
- npm imports

Happy to just lump together into 1 giant ball of mush.

src/commands/version.ts
```diff
@@ -0,0 +1,239 @@
...
+export interface ModuleVersion {
+	name: string;
+	version: string;
+	group: string;
+}
+
+/**
+ * Important information to be retrieved from a module's package.json
+ */
+export interface PackageDetails {
```
rishson: 👍

src/commands/version.ts
rishson: 👍

src/commands/version.ts
rishson: Would be more robust - e.g. `fs.stat` then `require` 👍

src/commands/version.ts
rishson: 👍

src/commands/version.ts
rishson: Thinking about it, all load fails result in returning default vals, so better to just allow the failure.

src/command.ts
```diff
@@ -39,13 +36,52 @@ export function initCommandLoader(searchPrefix: string): (path: string) => Comma
...
+ */
+export function createBuiltInCommandLoader(): (path: string) => CommandWrapper {
+
+	return function load(path: string): CommandWrapper {
+		const module = require(path);
+
+		try {
+			const command = convertModuleToCommand(module);
+			//  derive the name and group of the built in commands from the command itself (these are optional props)
+			const { name = '', group = '', description, register, run } = command;
```
rishson: Don't understand people's avoidance of redundant locals. Code is easier to debug and just as efficient. In this case, the destructuring was just to give defaults for optional properties because I think it looks better that returning an object with `name || '',` etc..

rishson:
Don't get the codecov. `grunt test` gives coverage of 99.57%, but codecov on `ci` says 97.87%.

src/commands/version.ts
rishson: Also, added a note to #73 to split this into separate files - so will address in separate PR.

tests/unit/commands/version.ts
```diff
@@ -0,0 +1,166 @@
...
+			assert.isTrue(false);   // not sure how to test a promise is not rejected - sinon-as-promised?
```
matt-gadd: why do you need this? if the promise rejects the test should fail.

tests/unit/commands/version.ts
```diff
@@ -0,0 +1,166 @@
...
+		});
+	});
+
+	function outputPrefix() {
```
matt-gadd: why aren't these at the top? weird to have utility functions at the bottom of the file.

tests/unit/loadCommands.ts
```diff
@@ -18,63 +29,92 @@ let consoleStub: SinonStub;
...
 		},
-		async 'Should search given paths for prefixed command files'() {
-			await loadCommands(yargsStub, config, loadStub);
-			assert.isTrue(loadStub.calledTwice);
-			const loadPath = loadStub.firstCall.args[0];
-			assert.isTrue(loadPath.indexOf(config.searchPaths[0]) > -1);
-			assert.isTrue(loadPath.indexOf(config.searchPrefix) > -1);
+		async 'Should successfully enumerate installed commands'() {
+			const installedPaths = await enumInstalledCommands(config);
+			assert.isTrue(installedPaths.length === 2);
```
matt-gadd: equal?

tests/unit/loadCommands.ts
```diff
@@ -18,63 +29,92 @@ let consoleStub: SinonStub;
...
+		},
+		async 'Should successfully enumerate builtin commands'() {
+			const builtInPaths = await enumBuiltInCommands(config);
+			assert.isTrue(builtInPaths.length === 2);   // includes invalid commands
```
matt-gadd: equal again?

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+ * Version checks are async calls to npm - so module repository dependant for now.
+ *
+ * @param {ModuleVersion[]} moduleVersions
+ * @returns {{name, version, group}[]}
+ */
+function areCommandsOutdated(moduleVersions: ModuleVersion[]): Promise<any> {
+	// convert [ModuleVersion] = [{commandPackageName: commandPackageVersion}]
+	let deps: {}[] = moduleVersions
+		.map((command) => {
+			let obj: {
```
matt-gadd: why `let`?

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+				version,
+				group
+			};
+		})
+		.filter((val) => {
+			// remove inbuilt commands or commands that dont have a valid version in the package.json
+			return val.version !== versionNoVersion && val.version !== INBUILT_COMMAND_VERSION;
+		})
+		.sort((a, b) => a.group > b.group ? 1 : -1);
+	return Promise.resolve(versionInfo);
```
matt-gadd: why is this a promise yet no async occurring?

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/**
+ * Iterate through a CommandsMap and retrieve the module details of each module referenced in the
+ * CommandsMap. The returned list is sorted in alphabetical order, by group.
+ *
+ * @param {CommandsMap} commandsMap
+ * @returns {{name, version, group}[]}
+ */
+function buildVersions(commandsMap: CommandsMap): Promise<any> {
```
matt-gadd: return type?

tests/unit/command.ts
```diff
@@ -83,6 +105,21 @@ registerSuite({
...
+			loader = command.createBuiltInCommandLoader();
+		},
+		'Should throw an error when attempting to load'() {
+			try {
+				commandWrapper = loader(getBuiltInCommandPath(true));
+				assert.fail(null, null, 'Should not get here');
+			}
+			catch (error) {
+				assert.isTrue(error instanceof Error);
+				assert.isTrue(error.message.indexOf('does not satisfy the Command interface') > -1);
```
matt-gadd: why can't we assert the full message?

src/registerCommands.ts
```diff
@@ -21,23 +17,23 @@ export interface YargsCommandNames {
...
+
 	const commandHelper = new CommandHelper(commandsMap, helperContext);
-	const helper = new Helper(commandHelper, yargs, helperContext);
+	const helper = new Helper(commandHelper, yargs, helperContext, commandsMap);
 
-	Object.keys(yargsCommandNames).forEach((group: string) => {
-		const commandNames = yargsCommandNames[group];
-		const groupDescription = getGroupDescription(commandNames, commandsMap);
-		const defaultCommand = <CommandWrapper> commandsMap.get(group);
+	yargsCommandNames.forEach((val, key, map) => {
```
matt-gadd: why are these now generically named? `val`, `key`?

src/index.ts
```diff
@@ -13,16 +14,34 @@ const pkg = <any> require(packageJsonFilePath);
...
+		const installedCommands = await loadCommands(installedCommandsPaths, installedCommandLoader);
+
+		// combine the inbuilt and installed commands - last in wins when keys clash
+		const commands = new Map([...installedCommands.commandsMap, ...builtInCommands.commandsMap]);
+		const yargsCommandNames = new Map([...installedCommands.yargsCommandNames, ...builtInCommands.yargsCommandNames]);
+
+		// register all the command with yargs and allow it to do the commands processing
+		registerCommands(yargs, commands, yargsCommandNames);
+	} catch (err) {
+		console.log(`Some commands are not available: ${err}`);
```
matt-gadd: If an error throws in the above, surely no commands would be registered?

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+export const versionCurrentVersion = `
+You are currently running dojo-cli {version}
+`;
+export const versionNoRegisteredCommands = `
+There are no registered commands available.`;
+export const versionNoVersion = yellow('package.json missing');
+export const versionRegisteredCommands = `
+The currently installed groups are:
+`;
+const INBUILT_COMMAND_VERSION = '__IN_BUILT_COMMAND__';
```
matt-gadd: could this not be a Symbol?

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/**
+ * Read information about a package/module, if available, or return default values.
+ *
+ * @param {string} packageDir The directory containing the package.json file.
+ * @returns {{name: (any|string), version: any}}
+ */
+function readPackageDetails(packageDir: string): PackageDetails {
+	let data: any = {};
```
agubler: could potentially use `Object.create(null);`?

src/commands/version.ts
agubler: but if `isBuiltInComment` is true, you do not even need to try to `require`, seems like all built in command will fail and fall into the catch.

So surely something like this would reduce the number of try/catch failures on require.

```
if (isBuiltInCommand(packageDir)) {
    data.version = INBUILT_COMMAND_VERSION;
} 
else {
    try {
        require(join(packageDir, 'package.json'));
    }
    catch (err) {
        // do nothing
    }
}
```

tests/unit/loadCommands.ts
rishson: 👍

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+	return new Promise((resolve, reject) => {
+		// we want to fetch the latest stable version for our devDependencies
+		david.getUpdatedDependencies(manifest, { dev: true, stable: true }, function (err: any, deps: {[dependencyName: string]: {stable: string }}) {
+			if (err) {
+				reject(err);
+			}
+			resolve(moduleVersions.map((command) => {
+				const canBeUpdated = deps[command.name];    // david returns all deps that can be updated
+				const versionStr = canBeUpdated ?
+					`${command.version} ${yellow(`(can be updated to ${deps[command.name].stable})`)}.` :
```
agubler: Is this covered by unit tests?

src/index.ts
agubler: Also not unit tested.

src/commands/version.ts
rishson: 👍 - old code

tests/unit/command.ts
```diff
@@ -28,6 +29,10 @@ function getCommandPath(prefix: string): string {
 	return `../tests/support/${prefix}-${testGroup}-${testName}`;
 }
 
+function getBuiltInCommandPath(invalid: boolean): string {
```
agubler: Can't you just have two constants and use them where ever you need rather than a function with a ternary?

tests/unit/loadCommands.ts
```diff
@@ -18,63 +29,92 @@ let consoleStub: SinonStub;
...
+		}
+	},
+	'unsuccessful enumeration': {
+		async 'Should fail to find installed commands that dont exist'() {
+			const cloneConfig = JSON.parse(JSON.stringify(config)); // trashy copy - better than post test reset
```
agubler: Can you not use `Object.assign({}, config);`?

src/commands/version.ts
rishson: 👍 - old code

src/registerCommands.ts
rishson: Always done generic args for iteration - probs too much lodash in my youth.

tests/unit/loadCommands.ts
matt-gadd: @agubler as long as the object isn't deep then sure

src/commands/version.ts
```diff
@@ -0,0 +1,234 @@
...
+}
+
+function run(helper: Helper, args: VersionArgs): Promise<any> {
+	const checkOutdated = args.outdated !== undefined;
+	if (checkOutdated) {
+		console.log('Fetching latest version information...');
+	}
+
+	return createVersionsString(helper.commandsMap, checkOutdated)
+		.then((data) => console.log(data));
```
rishson: I don't think the comment line is redundant as its trying to explain that builtin commands do something different to installed commands when it comes to assigning the `name` and `group` properties. This is in contrast to the corresponding comment in `initCommandLoader`: `//  derive the group and name from the module directory name, e.g. dojo-cli-group-name`

rishson: I also dislike promise resolutions that imply the data being passed e.g. `.then(console.log)` as these are harder to debug. I'll change, but still don't see what is gained by making the code as terse as possible, and removing opportunities to debug.

src/commands/version.ts
rishson: Was trying to keep the refactor small and not re-write where possible.
2nd point  - nope, needs a test for updateable deps - will add.

tests/unit/command.ts
rishson: 1st: yep - covering this last line is again a refactor, as we have already loaded the `cli` module in the tests. I mentioned in my comment for the PR why:
```
Getting code coverage from 99.87% to 100% will require introducing extra dependencies or re-writing how the cli main entry func works, so I have not done this.
```

src/commands/version.ts
agubler: Assume the comment comment is against the wrong initial comment?

I don't think `.then((data) => console.log(data))` is any clearer or any easier to debug as `.then(console.log)` personally. But you are creating one less anonymous function with the later which is good and not implicitly return the result of `console.log(data)` afaik

rishson: `.then((data) => console.log(data))` - I can set a debugger on this line and interrogate the value of `data` - where as with `.then(console.log)` I cannot (at least in my IDE) - hence my preference for the former.

agubler: Okay but I don't think we should implicitly be returning the result of `console.log` regardless.

Otherwise it is pretty much personal preference, ignoring the extra anonymous function that is created.

rishson: Changed to:
```
		.then(function(data) {
			console.log(data);
		});
```

Debuggable - and no void return from `console.log`.

agubler: You can actually do 

```ts
(data) => {
    console.log(data);
}
```

tests/unit/commands/version.ts
rishson: Done

tests/unit/commands/version.ts
Tomdye: @agubler pointed out that this has been changed now, so disregard my comment

src/index.ts
rishson: 👍

tests/unit/commands/version.ts
rishson: 👍

tests/unit/commands/version.ts
rishson: 👍

src/commands/version.ts
rishson: Actually was tested - who knew.

src/commands/version.ts
```diff
@@ -0,0 +1,240 @@
...
+function readPackageDetails(packageDir: string): PackageDetails {
+	let data: any = {};
+	// rather than add another prop to Command, declare the command to be builtin by setting its version
+	if (isBuiltInCommand(packageDir)) {
+		data.version = INBUILT_COMMAND_VERSION;
+	} else {
+		try {
+			data = require(join(packageDir, 'package.json'));
+		} catch (e) {
+			data.name = packageDir;
```
agubler: This is not covered by unit tests :(

src/index.ts
```diff
@@ -13,16 +13,34 @@ const pkg = <any> require(packageJsonFilePath);
...
+		const installedCommands = await loadCommands(installedCommandsPaths, installedCommandLoader);
+
+		// combine the inbuilt and installed commands - last in wins when keys clash
+		const commands = new Map([...installedCommands.commandsMap, ...builtInCommands.commandsMap]);
+		const yargsCommandNames = new Map([...installedCommands.yargsCommandNames, ...builtInCommands.yargsCommandNames]);
+
+		// register all the command with yargs and allow it to do the commands processing
+		registerCommands(yargs, commands, yargsCommandNames);
+	} catch (err) {
+		console.log(`Commands are not available: ${err}`);
```
agubler: not covered by unit tests :(

rishson: as mentioned - I'll do in a separate PR because its a bigish refactor, unless someone has a bright idea on testing branches in a module that auto-runs, e.g. 

```
function init(){...}
init();
```

rishson: 👍

src/commands/version.ts
rishson: 👍

agubler:
Nice @rishson, merge away at your own convenience 👍

PR #76, opened 2016-11-24
Pinned intern to 3.3.2

rishson:
Resolves typings issues with intern - so the `ci` build will pass.


rishson:
Ugh - I know. I really need a bash script that does `fetch upstream` before `push` on forks...

agubler:
this PR should **never** have all these commits! Pull from master and create a new branch from that?

agubler:
Let's sort that out before merging this PR

PR #80, opened 2016-11-25
Correctly structure the manifest

agubler:
Also fix issue that meant outdated check was run even when the `-o` flag wasn't passed


rishson:
👍

PR #84, opened 2016-11-30
Remove 'CommandsMap' from 'Command' interface

### LINKED ISSUE #83: CommandsMap is exposed in the Command interface - that bad mkay
😞 soz - leaky abstraction.

rishson:
Fixes #83.
`allCommands` is an async function that maintains a cache of all the commands (installed and built-in). This module is used by both the `cli` and the built-in `version` command.

In addition:
`MockModule` now able to mock modules that are relative to module under test (but not in same dir).
Typings for `sinon-as-promised`.


src/commands/version.ts
```diff
@@ -224,9 +225,10 @@ function run(helper: Helper, args: VersionArgs): Promise<any> {
 		console.log('Fetching latest version information...');
 	}
 
-	return createVersionsString(helper.commandsMap, checkOutdated)
-		.then(function(data) {
-			console.log(data);
+	return allCommands()
```
agubler: to avoid nesting, but I know you don't like it.

```ts
return allCommands()
    .then((commands) => {
	    return createVersionsString(commands.commandsMap, checkOutdated);
  	})
	.then(console.log);
```

tests/unit/allCommands.ts
```diff
@@ -0,0 +1,100 @@
...
+describe('AllCommands', () => {
+
+	let moduleUnderTest: any;
+	let mockModule: MockModule;
+	let sandbox: any;
+	let mockCommand: any;
+	let mockLoadCommands: any;
+
+	it('should run functions in order', () => {
+		describe('inner', () => {
```
agubler: inner is no good for identifying the test if there is a failure, i think you can drop it also... plus `it('should run functions in order', () => {` should probably be a `describe`

Issue #58, opened 2016-10-02
Help options misleading when more than one command for a group


agubler:
If there are more than one `command` for a `group` the options reflect just one of the commands when using the `dojo group -h`, this is misleading.

Example, the first help output shows the two commands available under `build` but the `options` are only valid for the `dojo` build.

**Command:** `dojo build -h`

``` shell
dojo build

Commands:
  dojo     Build dojo2 packages
  webpack  create a build of your application

Options:
  -h, --help     Show help                                             [boolean]
  -t, --type     type of build         [choices: "dev", "dist"] [default: "dev"]
  -v, --version  Show version number                                   [boolean]
```

**Command:** `dojo build webpack -h`

``` shell
dojo build webpack

Options:
  -h, --help     Show help                                             [boolean]
  -w, --watch    watch and serve
  -p, --port     port to serve on when using --serve                    [number]
  -v, --version  Show version number                                   [boolean]
```

**Command:** `dojo build dojo -h`

``` shell
dojo build dojo

Options:
  -h, --help     Show help                                             [boolean]
  -t, --type     type of build         [choices: "dev", "dist"] [default: "dev"]
  -v, --version  Show version number                                   [boolean]
```

rishson:
I don't think group help should show any options other than defaults `-h -v`, e.g.
`dojo -h` shows list of groups
`dojo someGroup` shows full help if only 1 command in the group, else shows the commands and default options :

```
dojo build

Commands:
  dojo     Build dojo2 packages
  webpack  create a build of your application

Options:
  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]
```

and default options should always go at the end:

```
dojo build webpack

Options:
  -w, --watch    watch and serve
  -p, --port     port to serve on when using --serve                    [number]

  -h, --help     Show help                                             [boolean]
  -v, --version  Show version number                                   [boolean]
```

rishson:
Actually, thinking about this some more, I think that `-h` when in a group should only show the options for that group, but when no group specified, e.g. `dojo -h` then show the non-group-specific options such as `-h -v`

agubler:
Addressed by #85

PR #85, opened 2016-12-05
Fix help so it displays per grouping

jdonaghue:
screen shot 2016-12-06 at 7 15 38 pm

Shows sub commands and the default command's options as well as global options
screen shot 2016-12-07 at 8 31 42 am
screen shot 2016-12-07 at 8 32 11 am

**Type:** bug

**Description:** 

Fix help display so that it displays options per grouping

**Related Issue:** #58

Please review this checklist before submitting your PR:

* [x] All contributors have signed a CLA
* [x] The code passes the CI tests
* [x] The PR increases or maintains the overall unit test coverage percentage
* [x] The code is ready to be merged

### DIFF
--- a/src/commands/version.ts
+++ b/src/commands/version.ts
@@ -1,7 +1,7 @@
 import { CommandsMap } from '../command';
 import { Helper, OptionsHelper } from '../interfaces';
 import { join } from 'path';
-import { Argv, Options } from 'yargs';
+import { Argv } from 'yargs';
 import { yellow } from 'chalk';
 import allCommands from '../allCommands';
 const david = require('david');
--- a/src/commands/version.ts
+++ b/src/commands/version.ts
@@ -1,5 +1,5 @@
 import { CommandsMap } from '../command';
-import { Helper } from '../interfaces';
+import { Helper, OptionsHelper } from '../interfaces';
 import { join } from 'path';
 import { Argv, Options } from 'yargs';
 import { yellow } from 'chalk';
@@ -122,7 +122,7 @@ function createOutput(myPackageDetails: PackageDetails, commandVersions: ModuleV
 	return output;
 }
 
-function register(options: (key: string, options: Options) => void): void {
+function register(helper: Helper, options: OptionsHelper): void {
 	options('o', {
 		alias: 'outdated',
 		describe: 'Output a list of installed commands and check if any can be updated to a more recent stable version.',
--- a/src/commands/version.ts
+++ b/src/commands/version.ts
@@ -1,7 +1,7 @@
 import { CommandsMap } from '../command';
 import { Helper } from '../interfaces';
 import { join } from 'path';
-import { Yargs, Argv } from 'yargs';
+import { Argv, Options } from 'yargs';
 import { yellow } from 'chalk';
 import allCommands from '../allCommands';
 const david = require('david');
@@ -122,14 +122,13 @@ function createOutput(myPackageDetails: PackageDetails, commandVersions: ModuleV
 	return output;
 }
 
-function register(helper: Helper): Yargs {
-	helper.yargs.option('o', {
+function register(options: (key: string, options: Options) => void): void {
+	options('o', {
 		alias: 'outdated',
 		describe: 'Output a list of installed commands and check if any can be updated to a more recent stable version.',
 		demand: false,
 		type: 'boolean'
 	});
-	return helper.yargs;
 }
 
 /**
--- a/src/commands/version.ts
+++ b/src/commands/version.ts
@@ -122,7 +122,7 @@ function createOutput(myPackageDetails: PackageDetails, commandVersions: ModuleV
 	return output;
 }
 
-function register(helper: Helper, options: OptionsHelper): void {
+function register(options: OptionsHelper): void {
 	options('o', {
 		alias: 'outdated',
 		describe: 'Output a list of installed commands and check if any can be updated to a more recent stable version.',
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -18,7 +18,7 @@ export type OptionsHelper = (key: string, options: Options) => void;
  */
 export interface Command {
 	description: string;
-	register(helper: Helper, options: OptionsHelper): void;
+	register(options: OptionsHelper): void;
 	run(helper: Helper, args?: Argv): Promise<any>;
 	name?: string;
 	group?: string;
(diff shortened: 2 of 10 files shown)


Tomdye:
This needs a test for the bug it's trying to fix.

Tomdye:
@jdonaghue have you been able to test this in the case that we have two commands for a given group? Please can you attach some screenshots to the PR of the output you are getting for a given input (as per previous CLI PR's) so than we can clarify the result.

jdonaghue:
@Tomdye I am working on the tests now and will add the screenshots shortly after that. Good suggestions, thanks!

Tomdye:
Im not sure the output is quite right @jdonaghue.

Given a situation where we had two build commands registered, `webpack` and `rjs`. 
When typing `dojo build -h` I would expect to see:

```
Commands:
   webpack
   rjs

Options:
   -h, --help   Show help
```

Only when typing `dojo build webpack -h` would I expect to see the options for webpack build showing:

```
Options:
   -w, --watch   Run as watch
   -f, --foo   Foo option
   -h, --help   Show help
```

Finally, if there is only one command registered under build, we would show the same output as above.

@agubler can you confirm if this is what you had in mind with the issue you raised?

jdonaghue:
@Tomdye I can certainly make it behave that way but not if we want to allow for a default command. The default command will run but not accept any options in that scenario. Currently, when the `build` command is registered we set the default commands (`dojo` in the example above) options. If we do this then those options will always display somewhere when `dojo build -h` is typed in. Either under the `Options` section or beneath a `dojo` section. Prior to this fix they were showing up under the `Options` section. This is because we are associating those options with the `build` command (as well as the `dojo` command. In order to get what you described above we would need to ditch the concept of a default command (while working with yargs anyway).

jdonaghue:
@Tomdye we could do a little conditional logic and if there is only one command not associate a group with the options that would allow for the 2nd scenario that you mentioned to work that way. But the 1st scenario will have to behave the way I implemented it if we are to keep the concept of a default command and not want only the first commands options to show when `dojo build -h` is entered.

Tomdye:
re. the default command. At the moment it is just the first command loaded for a given group, so it's not really a thing. We still need to work out if it's even something we will support when a group has multiple commands available and how we will deliver that. So for the time being I would think that given that information the output mentioned above should be possible.

src/interfaces.ts
```diff
@@ -16,7 +16,7 @@ export interface Helper {
  */
 export interface Command {
 	description: string;
-	register(helper: Helper): Yargs;
+	register(options: (key: string, options: Options) => void): void;
```
Tomdye: how would a command get access to the command helper now within it's `register` function?

src/commands/version.ts
```diff
@@ -122,14 +122,13 @@ function createOutput(myPackageDetails: PackageDetails, commandVersions: ModuleV
 	return output;
 }
 
-function register(helper: Helper): Yargs {
-	helper.yargs.option('o', {
+function register(options: (key: string, options: Options) => void): void {
```
Tomdye: this should be moved to a type: `(key: string, options: Options) => void` as it's used in the commands also

src/interfaces.ts
```diff
@@ -16,7 +16,7 @@ export interface Helper {
  */
 export interface Command {
 	description: string;
-	register(helper: Helper): Yargs;
```
Tomdye: how would a command get access to the `helper` now from its register function?

jdonaghue: I didn't think it was necessary any longer since the new `options` parameter will do the registration. I can always add it as a second (or first) param

src/commands/version.ts
jdonaghue: good idea!

jdonaghue:
OK cool if we don't need to support a default command then yes that format is possible. I will make the changes necessary to accommodate that. Thanks!

src/interfaces.ts
jdonaghue: I added back in `helper` and left `options` although `options isn't technically needed any longer since it isn't doing anything special to the options. I left it though so that we could easily add modifications in the future. Should I leave it or should I just remove and have the commands call `helper.yargs.options`?

jdonaghue:
I have made the desired changes and here is what the help output looks like now:

screen shot 2017-01-01 at 6 22 39 pm

screen shot 2017-01-01 at 6 23 06 pm

screen shot 2017-01-01 at 6 46 16 pm

src/interfaces.ts
Tomdye: I kinda liked how it was before you put `helper` back in, was a better abstraction away from `yargs` for each command.

jdonaghue: So should we just remove the helper from `register` and have it look like this then:

```
export interface Command {
        ...
        register(options: OptionsHelper)
        ...
}

I liked that too

Tomdye: I meant how the options function was passed in rather than `yargs`

Tomdye:
This still needs `defaultcommand` to be re-instated

agubler:
@jdonaghue are you still working on reinstating default command functionality?

jdonaghue:
Ok this is the best that I could do for the default command:
screen shot 2017-01-26 at 7 52 45 pm

Here is for dojo:
screen shot 2017-01-26 at 7 53 08 pm

Here is webpack:
screen shot 2017-01-26 at 7 53 23 pm

The reason for the default command help options looking the way they do is that command option registration occurs prior to command execution. So at the time when registering options we don't know how the cli is being called making it impossible to conditionally register option based on call signature. This combined with the fact that you must register options upfront when using `yargs` in order to allow them to be entered, restrict that only they may be entered and then show them in help makes it impossible to not show the default commands options when running help on that command (eg `dojo build -h`).

So we have a choice between two scenarios:
1. Just show the default options under the `Options` section (not indicating which command they are for.
2. Group the options under a descriptive name (as is done in the picture above) and showing them with some context.

I think what we would ideally want is to tell `yargs` about those options but not have them displayed in this scenario, but that is (as far as I can tell with the current version of `yargs@4.8.1`) impossible.

tomdye:
PR merge is dependant on: https://github.com/dojo/cli-create-app/pull/29 and https://github.com/dojo/cli-build/pull/26

jdonaghue:
@tomdye https://github.com/dojo/cli-create-app/pull/29 and https://github.com/dojo/cli-build/pull/26 are both failing only because this PR hasn't landed yet. Since they both depend on changes to `interfaces` in this PR but are pulling `. Should I just update those two PRs to point to the most recent commit from this PR? Or is it better to land this and publish it first and then update those two PR's with the newly published version?

agubler:
@jdonaghue we shall merge and publish these changes then update the dependant PRs to point to the updated CLI version.

Issue #44, opened 2016-09-21
Eject command


matt-gadd:
Our cli commands are meant to make the consumers code config free by internalising the config/config generation into the command itself. Given this, if a user needs more flexibility than the basic setup, we should allow them to "eject" those configs from the cli so they can use as a base for them to build on. This is an idea implemented in create-react-app.

The easiest way to implement this would just be to add an additional function to the command interface named `eject`, it is then up to the command to dump whatever config it has in files to the project.

Additional thoughts:
- Nearly all eject commands are going to want to add dependencies to the package.json, so we might want to implement a helper to do that.
- `dojo eject` seems like it should be an internal command to the cli. We don't have the concept of internal commands at the moment.

Tomdye:
Will this need to re-write the `index.html` to include script tags etc...?

matt-gadd:
@Tomdye no, the webpack config does that anyway.

jdonaghue:
@matt-gadd @Tomdye After looking at this for a little while I feel like this is a little more complex than it is for `create-react-app` or `enclave-eject`. Mostly because the `dojo cli` is designed to provide for many different commands that can potentially be totally unrelated and are unknown until the cli begins execution. Whereas with the previous two cli tools, they have knowledge upfront of what all of the configuration is that needs to be ejected. 

So that brings me to the following questions:
* Do we want `dojo eject` to behave somewhat similarly and just eject all configs from all registered commands as a blanket operation?
    * Basically, by discovering each command (as it does now) and then calling into an `eject` method defined within each command.
* Or should it be more granular and allow for only a certain commands configurations to be ejected based on what is specified when calling `dojo eject`? 
    * If so what should that look like in terms of execution? Something like `dojo eject build webpack` which would basically call an `eject` method built into the `dojo-cli-build-webpack` module?

Either of these make me a little nervous. Mostly because there isn't an overarching control of dependencies within the `dojo cli` like there is in the other two tools. The other two tools are (somewhat) atomic in nature and as a result don't have the risks for collisions etc that `dojo cli` does.

What are your thoughts?

jdonaghue:
One thing we could do to mitigate the risk of collisions is to make each command module not be responsible for its own ejection, but instead just return a collection of configuration/dependencies to the base cli which would then after collecting all of these execute the eject after considering for any potential conflicts.

agubler:
[...]

PR #89, opened 2016-12-15
Eject command - refs #44

jdonaghue:
Putting this out now for initial review. The basic idea is that each command will optionally implement an `eject` method. The base cli will call into that `eject` method with two helpers `npm` and `files` which respectively add to the current projects `package.json` and copies important files over to the current project. 

`npm` is capable of assigning `dependencies`, `devDependencies`, and `scripts` to the current projects `package.json`. It will run `npm install` after it has finished assignments but only if there were no collisions.

`files` takes an array of fully-qualified file paths and copies them into the current project creating the necessary folders to do so, as long as there are no collisions.

Keep in mind this is still a WIP and needs better collision checking/messaging as well as a prompt to ensure the user understands the finality of this commands execution.

It can be run in the following way:

* `dojo eject` - will run through all commands and attempt to eject if the method is defined. Right now it will fail if a command doesn't have an `eject` defined but that should probably only be the case when there is a specific command specified and not for this generic execution pattern.
* `dojo eject -g groupName` - will run through all commands within a group and attempt to eject if the method is defined. Again it will fail if a command doesn't define an `eject` method, but that behavior should probably be reserved for when a specific command is specified.
* `dojo eject -g groupName -c commandName` - will try to eject this command. If an `eject` method is not implemented on the command it will fail.

Also, tests are still needed.


src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+		parts.pop();
+		if (parts.length) {
+			folders.push(join(...parts));
+		}
+	});
+
+	// create those paths in the current project
+	folders.forEach((folder) => {
+		console.log(`creating folder: ${join(appPath, folder)}`);
+		fs.mkdirSync(join(appPath, folder));
```
Tomdye: would `mkdirp` be a better fit? Allows you to create a full path rather than one folder at a time.

Tomdye: `import { mkdirsSync } from 'fs-extra';` is what is used in `cli-create-app`

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
+import * as fs from 'fs';
+import { Helper, NpmPackage } from '../interfaces';
+import { join } from 'path';
+import { Yargs, Argv } from 'yargs';
+import { yellow } from 'chalk';
+import allCommands from '../allCommands';
+const spawnSync = require('cross-spawn');
```
Tomdye: Looking at the docs for `cross-spawn` it returns `spawn` not a `spawnSync` function.
From the docs:

```
var spawn = require('cross-spawn');
 
// Spawn NPM asynchronously 
var child = spawn('npm', ['list', '-g', '-depth', '0'], { stdio: 'inherit' });
 
// Spawn NPM synchronously 
var results = spawn.sync('npm', ['list', '-g', '-depth', '0'], { stdio: 'inherit' });
```

jdonaghue: good catch! I will correct that.

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	// copy over files to the current project
+	console.log(yellow(`copying files into current project at: ${appPath}`));
+	files.forEach(function(file) {
+		const newPath = file.replace(longestCommonPath, '');
+		console.log(`->->copying ${yellow(file)} to the project which will now be located at: ${join(appPath, newPath)}`);
+		if (fs.existsSync(join(appPath, newPath))) {
+			throw Error(`File already exists: ${join(appPath, newPath)}`);
+		}
+		const content = fs.readFileSync(file, 'utf8').trim() + '\n';
```
rishson: Why the read, trim, append newline, rather than `fs.renameSync(file, join(appPath, newPath))`?

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Helper - add npm configuration to package.jsonn
+ * @param {NpmPackage} package
+ * @returns {void}
+ */
+function handleNpmConfiguration(pkg: NpmPackage): void {
+	const appPackage = require(join(appPath, 'package.json'));
+
+	Object.keys(pkg.devDependencies).forEach(function (dependency) {
+		console.log(`->->adding ${yellow(dependency)} to devDependencies`);
```
Tomdye: I don't personally like the `->->` format, but that's just a preference. What does the `create-react-app` output when you run its `eject` command for example.

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+
+	Object.keys(pkg.dependencies).forEach(function (dependency) {
+		console.log(`->->adding ${yellow(dependency)} to dependencies`);
+		appPackage.dependencies[dependency] = pkg.dependencies[dependency];
+	});
+
+	Object.keys(pkg.scripts).forEach(function (script) {
+		console.log(`->->adding ${yellow(script)} to scripts`);
+		if (appPackage.scripts[script]) {
+			throw Error(`package script ${yellow(script)} already exists`);
```
Tomdye: Not sure if this should be an error, haven't looked ahead to see if this error is caught somewhere, but more likely we'd just want to display a warning rather than failing.

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+		}
+		appPackage.scripts[script] = pkg.scripts[script];
+	});
+
+	console.log(yellow('running npm install...'));
+	spawnSync('npm', ['install'], { stdio: 'inherit' });
```
Tomdye: Can you run `npm install` the same as we do here please?

https://github.com/dojo/cli-create-app/blob/master/src/npmInstall.ts

Perhaps it should be moved into the core cli and accessed via a helper?

agubler: feels like we'd be better following the pattern that `cli-create-app` has used, in that it silences the verbose output but displays a small spinner whilst the installations happens.

https://github.com/dojo/cli-create-app/blob/master/src/npmInstall.ts

src/commands/eject.ts
Tomdye: i'd used it in `cli-create-app` so it jogged my memory

src/commands/eject.ts
agubler: haha jinx @Tomdye

Tomdye: We both added the same comment here @agubler great minds right? Do you think that could be moved out of `cli-create-app` and into `cli` itself accessible via a helper for consistency?

agubler: haha jinx @Tomdye

src/commands/eject.ts
jdonaghue: `create-react-app` is very similar but doesn't use a `->->` I can get rid of that.

src/commands/eject.ts
agubler: @Tomdye very possibly...

src/commands/eject.ts
rishson: If you add `fs-extra` then it has a `copySync` so you could replace 102 -103.

src/commands/eject.ts
jdonaghue: I wasn't sure if it should be an error or not either. However, if the user was expecting the exact same functionality as the cli provided and a script didn't make it because of a collision it could be a real frustration if they overlooked the warning and discovered it only while debugging later.

src/commands/eject.ts
jdonaghue: haha good idea guys. I should have looked at `cli-create-app` first. I will change this

src/commands/eject.ts
jdonaghue: nice @rishson! Thanks for the tip I will do that.

src/commands/eject.ts
```diff
@@ -0,0 +1,151 @@
...
+	// TODO: add in prompt
+	return allCommands()
+		.then((commands) => {
+			const toEject = [ ...commands.commandsMap ]
+				.filter(([ , command ]) => {
+					return (!args.group || args.group === command.group) &&
+						(!args.command || args.command === command.name);
+				});
+
+			if (!toEject.length) {
```
jdonaghue: yeah that is a good idea. I know this current logic is a bit aggressive in terms of throwing errors. I will refine that today when I add in the prompt and make other changes that you guys recommended.

agubler: perfect!

Gruntfile.js
```diff
@@ -1,12 +1,35 @@
...
 				expand: true,
-				cwd: '.',
-				src: 'tests/**/*.json',
-				dest: '<%= tsconfig.compilerOptions.outDir %>'
+				cwd: 'src',
+				src: staticFiles,
+				dest: '<%= distDirectory %>'
 			}
-		}
+		},
```
Tomdye: trailing comma

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+	files.forEach(function(filePath) {
+		const newPath = pathResolve(filePath).replace(longestCommonPath, '');
+		if (existsSync(join(appPath, newPath))) {
+			throw Error(`File already exists: ${join(appPath, newPath)}`);
+		}
+	});
+
+	// create those folders within the current project
+	folders.forEach((folder) => {
+		console.log(`creating folder: ${yellow(join(appPath, folder))}`);
```
Tomdye: could assign result of `join(appPath, folder)` to a `const` as it's used twice

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+const appPath = pkgDir.sync(process.cwd());
+
+/**
+ * Helper - add npm configuration to package.jsonn
+ * @param {NpmPackage} package
+ * @returns {void}
+ */
+async function handleNpmConfiguration(pkg: NpmPackage): Promise<void> {
+	const appPackage = require(join(appPath, 'package.json'));
+
```
Tomdye: You could use de-structuring defaults and an arrow function here
``` typescript
const { devDependencies = {}, dependencies = {}, scripts = {} } = pkg;
Object.keys(devDependencies).forEach((devDependency) => {
   // ... do stuff
});
```

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+	Object.keys(pkg.devDependencies || {}).forEach(function (dependency) {
+		console.log(`	adding ${yellow(dependency)} to devDependencies`);
+		appPackage.devDependencies = appPackage.devDependencies || {};
```
Tomdye: should be doing this outside the loop and if it does not exist, bypass this loop and simply assign the `pkg.devDependencies` instead of `{}`.

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+			console.warn(`${red('WARN')}    dependency ${dependency} already exists at version '${appPackage.dependencies[dependency]}' and will be overwritten by version '${pkg.dependencies[dependency]}'`);
+		}
+		appPackage.dependencies[dependency] = pkg.dependencies[dependency];
+	});
+
+	Object.keys(pkg.scripts || {}).forEach(function (script) {
+		console.log(`	adding ${yellow(script)} to scripts`);
+		appPackage.scripts = appPackage.scripts || {};
+		if (appPackage.scripts[script]) {
+			throw Error(`package script ${yellow(script)} already exists`);
```
Tomdye: Should this be an error or just a warning that we won't be overwriting the script?

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+		}
+		appPackage.scripts[script] = pkg.scripts[script];
+	});
+
+	if (pkg.dependencies || pkg.devDependencies || pkg.scripts) {
```
Tomdye: with above de-structuring approach, this would be `if (dependencies || devDependencies || scripts)`

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+		await npmInstall();
+	}
+}
+
+/**
+ * Helper - copy files to current project root
+ * @param {string[]} files - an array of fully-qualified file paths
+ * @returns {void}
+ */
+function copyFiles(files: string[]): void {
```
Tomdye: I think that a lot of this could likely be removed and something like `ensureDir` from `fs-extra` could be utilised instead?

src/commands/eject.ts
```diff
@@ -0,0 +1,187 @@
...
+		name: 'eject',
+		message: 'Are you sure you want to eject (this is a permanent operation)?',
+		default: false
+	}).then((answer: { eject: boolean }) => {
+		if (!answer.eject) {
+			throw Error('Aborting eject');
+		}
+		return allCommands()
+			.then((commands) => {
+				const map: { [name: string]: boolean } = { 'eject/': true, 'version/': true };
```
Tomdye: can you explain what is this doing please?

src/commands/eject.ts
jdonaghue: thats a great idea - I will do that, thanks!

src/commands/eject.ts
jdonaghue: Its keeping track of commands that have been ejected so that they don't get ejected twice. Additionally, it filters out `eject` and `version` as we don't ever want to eject those.

src/commands/eject.ts
jdonaghue: I can make it a warning

jdonaghue:
@Tomdye thanks for the review! I will make these changes today

src/commands/eject.ts
jdonaghue: Actually creating the directories is only a small part of what this is doing. Most of the code here is for dealing with translating file paths to be relative to the current project root. 

This helper receives a list of fully qualified file paths that could be located anywhere (but are most likely within the users projects node_modules directory or the global node package directory) and needs to translate them into relative to the current users project root. So I don't think it will reduce any code. 

However, `ensureDirSync` is probably safer to use than `mkdirsSync`. I will switch to that thanks for the suggestion!

agubler:
@jdonaghue how are these updates coming?

src/commands/eject.ts
jdonaghue: didn't end up doing this due to an issue in istanbul/escodegen not being able to handle defaults https://github.com/gotwarlost/istanbul/issues/613

src/commands/eject.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+async function handleNpmConfiguration(pkg: NpmPackage): Promise<void> {
+	const appPackagePath = join(appPath, 'package.json');
+	const appPackage = require(appPackagePath);
+
+	appPackage.devDependencies = appPackage.devDependencies || {};
+	appPackage.dependencies = appPackage.dependencies || {};
+	appPackage.scripts = appPackage.scripts || {};
+
+	Object.keys(pkg.devDependencies || {}).forEach((dependency) => {
```
tomdye: I believe that these could be better implemented as a `reduce`

jdonaghue: @tomdye Reduce sounds good. I will do that. As far as the destructuring, do you want me to destructure even though defaults can't be used because of that bug in istanbul that removes the defaults while adding its coverage lines https://github.com/gotwarlost/istanbul/issues/613? I opted to not because because I would still need to do the `|| {}`, but if you would prefer that I can definitely add it in?

jdonaghue: @tomdye I have made changes to use `reduce`.

jdonaghue:
@tomdye I have changed the npm helper to use reduce

agubler:
@jdonaghue this now has a conflict in `src/interfaces.ts`, I would imagine as a result of #85 having been merged.

jdonaghue:
OK @agubler thanks I will resolve the conflict today.

tests/unit/bin/dojo.ts
```diff
@@ -0,0 +1,33 @@
...
+		mockYargs.ctor.alias = sandbox.stub().returns(mockYargs.ctor);
+		mockYargs.ctor.strict = sandbox.stub().returns(mockYargs.ctor);
+		moduleUnderTest = mockModule.getModuleUnderTest();
+	});
+
+	afterEach(() => {
+		mockModule.destroy();
+	});
+
+	it('should print help', () => {
```
agubler: Are we missing the test here? Or is this included accidentally?

Issue #88, opened 2016-12-14
cli should be able to store basic config on a per command basis


matt-gadd:
Although the cli commands are designed to be mostly config free, it seems like in quite a few cases some basic preferences are desirable. An example of this would be the `locale` in cli-build, or the `functional` or `unit` test modes in `cli-test`

A command would be able to persist preferences for itself via a cli setter api - which would be then stored in a `.dojorc` at the project level. The command would receive preferences on load (and/or a getter api)

agubler:
Done via #90

PR #90, opened 2017-01-09
Cli configuration helpers

jdonaghue:
Adds a `ConfigurationHelper` that provides two methods `save` and `get` for supporting configuration within cli commands. refs #88

### DIFF
--- a/src/ConfigurationHelper.ts
+++ b/src/ConfigurationHelper.ts
@@ -1,0 +1,52 @@
+import { existsSync, writeFile } from 'fs';
+import { join } from 'path';
+import { red } from 'chalk';
+import { ConfigurationHelper } from './interfaces';
+const pkgDir = require('pkg-dir');
+
+const appPath = pkgDir.sync(process.cwd());
+
+/**
+ * ConfigurationHelper class which is passed into each command's run function
+ * allowing commands to persist and retrieve .dojorc config object
+ */
+export default class implements ConfigurationHelper {
+	/**
+	 * persists configuration data to .dojorc
+	 * checks for collisions
+	 * 
+	 * @param config - the configuration to save
+	 * @returns Promise - this an indicator of success/failure
+	 */
+	async save(config: any = {}): Promise<any> {
+		const dojoRc = this.get() || {};
+		Object.keys(config).forEach((key) => {
+			if (key in dojoRc) {
+				throw Error(`${red('ERROR')} .dojorc already contains a '${key}' property`);
+			}
+			dojoRc[key] = config[key];
+		});
+		return new Promise((resolve, reject) => {
+			writeFile('.dojorc', JSON.stringify(dojoRc), { flag: 'wr' }, (error: Error) => {
+				if (error) {
+					reject(error);
+					return;
+				}
+				resolve(true);
+			});
+		});
+	};
+
+	/**
+	 * retrieves the configuration object from the file system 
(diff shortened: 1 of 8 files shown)


src/ConfigurationHelper.ts
```diff
@@ -0,0 +1,52 @@
...
+	async save(config: any = {}): Promise<any> {
+		const dojoRc = this.get() || {};
+		Object.keys(config).forEach((key) => {
+			if (key in dojoRc) {
+				throw Error(`${red('ERROR')} .dojorc already contains a '${key}' property`);
+			}
+			dojoRc[key] = config[key];
+		});
+		return new Promise((resolve, reject) => {
+			writeFile('.dojorc', JSON.stringify(dojoRc), { flag: 'wr' }, (error: Error) => {
```
rishson: Should you join `appPath` to filename otherwise the `get` is from package root, but the `save` is to `cwd`?

jdonaghue: good catch! I will fix this

src/ConfigurationHelper.ts
```diff
@@ -0,0 +1,52 @@
...
+				if (error) {
+					reject(error);
+					return;
+				}
+				resolve(true);
```
jdonaghue: good point I will get rid of the boolean

src/ConfigurationHelper.ts
```diff
@@ -0,0 +1,52 @@
...
+	 * persists configuration data to .dojorc
+	 * checks for collisions
+	 * 
+	 * @param config - the configuration to save
+	 * @returns Promise - this an indicator of success/failure
+	 */
+	async save(config: any = {}): Promise<any> {
+		const dojoRc = this.get() || {};
+		Object.keys(config).forEach((key) => {
+			if (key in dojoRc) {
```
matt-gadd: By default we should probably scope this to per command, ie: the .dojorc would read:
```json
{
    "commands": {
        "cli-build-webpack": {
            "locale": "en"
        },
        "cli-test-intern": {
            "default": "functional"
        }
    }
}
```
that way you can't have a collision, and we also know what specifically wrote to what. In the future we can potentially have some global config, but it seems that there isn't that much overlap in configuration between commands so far.

src/interfaces.ts
```diff
@@ -1,5 +1,10 @@
 import { Argv, Yargs } from 'yargs';
 
+export interface ConfigurationHelper {
+	save(config: any): Promise<any>;
```
matt-gadd: it feels a bit inconsistent to have a `save` that is async, yet the `get` is sync?

jdonaghue: Should they both be async or sync? I could go either way.

src/ConfigurationHelper.ts
jdonaghue: This has been addressed

src/ConfigurationHelper.ts
jdonaghue: This has been addressed

src/ConfigurationHelper.ts
jdonaghue: This has been addressed

src/interfaces.ts
matt-gadd: your call 😄

jdonaghue: I went async, that felt a little better to me. :)

PR #93, opened 2017-01-13
Support scoped @dojo cli commands

### LINKED ISSUE #92: Support @dojo scoped commands
We should support loading commands from @dojo scoped dependencies as well as dependencies prefixed with `dojo-cli`.

agubler:
**Type:** feature

**Description:**

Support `@dojo` scoped dependencies.

Resolves #92

### DIFF
--- a/tests/unit/command.ts
+++ b/tests/unit/command.ts
@@ -9,8 +9,8 @@ const expectedEsModuleCommand = require('intern/dojo/node!../support/esmodule-pr
 const testGroup = 'foo';
 const testName = 'bar';
 const testSearchPrefixes = [ 'test-prefix' ];
-const testEsModuleSearchPrefix = [ 'esmodule-prefix' ];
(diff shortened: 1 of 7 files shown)


tests/unit/command.ts
```diff
@@ -8,9 +8,9 @@ const expectedEsModuleCommand = require('intern/dojo/node!../support/esmodule-pr
 
 const testGroup = 'foo';
 const testName = 'bar';
-const testSearchPrefix = 'test-prefix';
-const testEsModuleSearchPrefix = 'esmodule-prefix';
-const testEsModuleFailSearchPrefix = 'esmodule-fail';
+const testSearchPrefixes = [ 'test-prefix' ];
```
dylans: Any reason this one was pluralized but not the two after (given the two after were also turned into arrays)? I think it makes sense, just wanted to verify.

agubler: I just missed them 👍

rishson:
👍

Issue #45, opened 2016-09-21
Support multiple groups for a command


matt-gadd:
Currently a command can only provide for a single group and that group is decided by part of its package name ie `dojo-cli-{groupname}-{command}`. 

it would be nice if in the command interface you can specify multiple groups or aliases. This is needed to support the `dev` task, which essentially desugars to the `build` task with some different arguments.

jdonaghue:
So I have implementing a new interface that looks like this:

```
export interface Alias {
	name: string;
	description?: string;
	register?(helper: Helper): Yargs;
}
```

and the `Command` interface will have a new optional property that looks like this:

```
alias?: Alias[]|Alias;
```

In `loadCommands` I will consult this new `alias` property and associate the command for each `alias` provided. This means that in `registerCommands` we will still use the default `register` method for the default `group` (the one obtained from the package name), but will additionally have to, for each alias, run the alias specific `register` or the default `register` if there isn't one. Finally, we will do a similar thing with descriptions as we may want each alias to have its own description.

If everyone agrees with this, I would like to have the help PR #85 to land first so that we reduce conflicts with registering options for help

Tomdye:
@jdonaghue I think this is the right direction but having a register for it is misleading as it's not going to have it's own run function, plus we need to have the ability to provide hardcoded options.
I believe we need to have an interface like this:

``` typescript
export interface Alias {
   name: string;
   description: string;
   options: AliasOptions[]
} 

export interface AliasOptions {
   option: string;
   value?: (string | boolean | number)[]
}
```

With that in mind, if for example `dojo dev` aliased `dojo build -w` (which it likely will), we would not want to see the `-w` option when we run `dojo dev -h`, only the remaining `-p` option that `dojo build` specifies.

jdonaghue:
I think this makes a lot of sense @tomdye. It would be good to land #85 first as some of the changes to the `registerCommand` module needed for that ticket will be needed for this ticket as well. The reason those two projects (`cli-create-app` and `cli-build`) are failing CI is because they depend on the changes in #85 to be landed so they can pull the `OptionsHelper` type correctly from `interfaces`.

jdonaghue:
@tomdye I think I might actually just rebase this PR against the PR for #85 so that I can get this one written as I think #85 is close enough now that not much will change in terms of signatures etc.

tomdye:
Looks like this isn't fully working.
[...]

PR #94, opened 2017-01-29
Create aliases for commands

jdonaghue:
refs #45 add ability for aliasing commands (eg. `dojo dev` is the same as `dojo build dojo -w`)

### DIFF
--- a/src/registerCommands.ts
+++ b/src/registerCommands.ts
@@ -25,19 +25,8 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 
 	yargsCommandNames.forEach((commandOptions, commandName) => {
 		const groupDescription = getGroupDescription(commandOptions, commandsMap);
-		const defaultCommand = <CommandWrapper> commandsMap.get(commandName);
-		const defaultCommandAvailable = !!(defaultCommand && defaultCommand.register && defaultCommand.run);
 		const reportError = (error: Error) => console.error(chalk.red.bold(error.message));
 		yargs.command(commandName, groupDescription, (yargs: Yargs) => {
-			if (defaultCommandAvailable) {
-				defaultCommand.register((key: string, options: Options) => {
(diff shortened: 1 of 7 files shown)


agubler:
@jdonaghue There are conflicts here that need resolving here also.

src/registerCommands.ts
```diff
@@ -58,6 +65,33 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
 				return defaultCommand.run(helper, argv).catch(reportError);
 			}
 		});
+
+		// Now handle aliases
```
tomdye: This comment isn't really adding anything. It would be better to move this functionality into it's own appropriately named function and call that instead. The same likely goes for the functionality to register commands above.

src/registerCommands.ts
```diff
@@ -58,6 +65,33 @@ export default function(yargs: Yargs, commandsMap: CommandsMap, yargsCommandName
...
+						(aliasYargs: Yargs) => {
+							register((key: string, options: Options) => {
+								if (!aliasOpts || !aliasOpts.some((option) => option.option === key)) {
+									aliasYargs.option(key, options);
+								}
+							});
+							return aliasYargs;
+						},
+						(argv: Argv) => {
+							alias.options && alias.options.forEach((option) => {
```
tomdye: i think this should be a `reduce`

agubler:
Still a conflict on `src/interfaces` and a few missing branches of tests coverage.

jdonaghue:
@agubler I have the conflicts resolved locally, but didn't have time yesterday to add to the coverage. I will get to it tonight/tomorrow most likely. Sorry, for the delay.

agubler:
@jdonaghue no problem at all! Wasn't sure if you were aware 😄

jdonaghue:
@agubler @tomdye The test coverage has been enhanced and all conflicts are now resolved, thanks!

Issue #95, opened 2017-02-01
cli should not error if no commands have implemented eject


agubler:
Currently `eject` throws an error if no commands have implemented it, we should either not throw an error and report that now commands have implemented or not register the command if there are no available commands to eject,

PR #99, opened 2017-02-04
Don't error when nothing to do

jdonaghue:
issue #95 

* Do not throw error when nothing to do
* When no commands have `eject` method, message that there is nothing to do

### DIFF
--- a/src/commands/eject.ts
+++ b/src/commands/eject.ts
@@ -1,10 +1,10 @@
 import { existsSync, writeFileSync } from 'fs';
 import { copySync, ensureDirSync } from 'fs-extra';
 import { join, resolve as pathResolve } from 'path';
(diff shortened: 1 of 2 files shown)


src/commands/eject.ts
```diff
@@ -171,12 +170,14 @@ function run(helper: Helper, args: EjectArgs): Promise<any> {
 					if (args.group && args.command) {
 						throw Error(`command ${args.group}/${args.command} does not exist`);
 					}
-					throw Error('nothing to do');
+					console.log('nothing to do');
```
agubler: Do you think we should be a bit more descriptive, i.e. `console.log('no commands have implemented eject')` or something similar?

jdonaghue: @agubler I have updated this message. Thanks!

Issue #97, opened 2017-02-02
Unable to find cli commands installed globally


agubler:
Seems that support for globally installed commands has broken and they are just not detected as available.

`npm install @dojo/cli -g`
`npm install @dojo/cli-create-app -g`

run `dojo version` and only the cli version is returned or run `dojo` and there is no entry help entry for the command.

PR #100, opened 2017-02-05
refs #97 look globally for cli commands

jdonaghue:


src/config.ts
```diff
@@ -12,7 +12,7 @@ export default {
 	searchPaths: [
 		'node_modules',
 		join(packagePath, '..'),
-		join(packagePath, 'node_modules')
+		join(dirname(process.execPath), '..', '/lib/node_modules')
```
agubler: @jdonaghue I think we've changed the wrong path here, because this has stopped command dependencies of the CLI being loaded. I believe the change should have been to the search path above it, `join(package, '..')` because since we moved to `@dojo` dependencies it meant that the `cli` was installed one directory deeper and therefore we needed to start the search path one directory further up.

```ts
join(packagePath, '../..'),
```

Obviously currently do not have good test coverage for either of those to scenarios....

jdonaghue:
@agubler Good catch! This has been updated to put back `join(packagePath, 'node_modules')`

src/config.ts
```diff
@@ -11,7 +11,7 @@ export type Config = {
 export default {
 	searchPaths: [
 		'node_modules',
-		join(packagePath, '..'),
+		join(dirname(process.execPath), '..', '/lib/node_modules'),
```
agubler: just a quick Q, if we dropped the `..` could we also drop the `lib` prefix to `node_modules`?

jdonaghue: No, because `dirname(process.execPath)` resolves to the `bin` directory where the `node` executable is located. So we have to go up one directory and then down into the `lib` directory which is a sibling of `bin` to get at `node_modules`.

matt-gadd: I don't think this is portable. From the npm docs:
```
Global installs on Unix systems go to {prefix}/lib/node_modules. Global installs on Windows go to {prefix}/node_modules (that is, no lib folder.)
```
Can we not use `__dirname` which is where the module is located, and come up from there?

jdonaghue: @matt-gadd Good catch! 

I was trying to go for an implementation that would work regardless of where the cli is installed. I realize that's not a strict requirement but I just figured that the only reason we would need to force install it globally is because of module searching. I could pretty easily conditionally include `lib` for unix systems and not for windows. 

Or we could just do the `__dirname` thing and require it to be installed globally. I personally prefer the first approach, but either way really works so whatever you guys think is best I will change to that.

Thanks!

jdonaghue: another options is `childProcess.execSync('npm root -g').toString()` but I like going from the `execPath` a little better for some reason.

matt-gadd: I guess my question is, what is the intended behaviour if you don't install the cli globally? I think there is probably an equal argument that if it's not installed globally we shouldn't be searching for commands globally either. And instead just search where the cli was installed (which is what coming up from __dirname would do anyway).

Either way I'm not a big fan of OS specific sniffs, because although the `lib` directory is unix specific, I don't doubt there are exotic cases that break that rule.

jdonaghue: That's a fair point about OS specific sniffs and I agree there isn't much of a use case for not installing globally. I suppose I was looking a bit too literally at the request to search globally and in practice there isn't much need for going beyond two levels up from `__dirname` in terms of what 'global' means.

jdonaghue:
@agubler @matt-gadd I have changed this so it goes up from `__dirname` now.

Issue #105, opened 2017-02-15
Drop `WARN - skipping` output from eject command


tomdye:
Currently the eject command outputs a bright red `WARN` message for every command that does not implement `eject`. This could be misleading to a user as there is no real error here. For example, the `cli-create-app` command will never have an eject so the user will always be shown an error here for something that is an author time decision, not a run time error.

Issue #107, opened 2017-02-15
Eject command does not allow for custom output directories


tomdye:
Presently the outputted files for each command ejected are written to the same location they originated from in the client project. We should look at allowing a target directory to be specified.

tomdye:
Have addressed this in #108 . All ejected files are put into a `command-name` folder to avoid clashes

Issue #106, opened 2017-02-15
Eject command will likely fail for multiple commands


tomdye:
Eject command builds `package.json` and runs `npm install` for every command it ejects one-by-one. The `npm install` command is an async process and nothing is currently waiting for it.

The eject command should ideally build up a list of files to copy and a combined package.json file before running `npm install` and `copy files` once only at the end. This will avoid making changes to the file system etc... before checking for clashes / issues.

Issue #104, opened 2017-02-15
Eject commands are not available in the command wrapper


tomdye:
The command wrapper that's used to store loaded commands currently does not expose the `eject` command. This needs to be changed to allow `eject` to work.

PR #108, opened 2017-02-15
Updates to get Eject working with real commands

tomdye:
### Updates to `eject` command internals.

- dropped `scripts` support, we don't use them.
- using return of `eject` rather than providing npm and copy files functions to the eject command.
- only running npm at the end
- copying files to directory named after command name
- added `path` to copy response to make relative file naming easier
- added `eject` to command wrapper
- added log output to say which command is being ejected
- removed `WARN` for commands that don't support `eject`
- made npm package contents optional
- changed `group/name` output to `group-name` to match command filenames
- updated tests

screenshot 2017-02-17 12 59 12

Fixes: #105 #104 #106 #107 
Blocks: https://github.com/dojo/cli-build/issues/58


src/commands/eject.ts
```diff
@@ -142,7 +26,20 @@ function register(options: OptionsHelper): void {
 	});
 }
 
-function run(helper: Helper, args: EjectArgs): Promise<any> {
+function copyFiles(commandName: string, { path, files }: FileCopyConfig): void {
+	if (path && files && files.length > 0) {
+		files.forEach((fileName) => {
+			const cwd = process.cwd();
+			const sourcePath = resolve(path, fileName);
+			const destPath = resolve(cwd, commandName, fileName);
```
jdonaghue: @tomdye I wrote the previous code assuming that the 'from' files could be within a nested directory structure. This doesn't appear to support that (although I may just be missing it). It looks like this is going to just move everything directly into `.../<commandName>/`. Is that not a thing that we should support?

src/commands/eject.ts
```diff
@@ -153,43 +50,59 @@ function run(helper: Helper, args: EjectArgs): Promise<any> {
...
 						}
 					});
-					if (!performedEject) {
-						console.log('no commands have implemented eject');
+				}
+				else if (args.group && args.command) {
+					throw Error(`command ${args.group}-${args.command} does not exist`);
+				}
+
+				if (commandsToEject) {
```
jdonaghue: @tomdye nice! This is a much better place to do the npm install. It didn't occur to me that it could potentially have conflicts across commands. Good catch!

src/commands/eject.ts
tomdye: The idea is that you specify the path to your command as well as relative paths to files such that you don't have to work that all out in the cli as you did previously. Each file is then relative to that path so that we can have nested structures. Chatting with @agubler we decided that putting each commands exported files into a `commandName` folder would avoid file clashes etc...

src/commands/eject.ts
tomdye: No worries, this way we don't upset the layout / formatting of the `package.json` file and the actual output of the resulting `npm install` call will provide the user with any info about package clashes etc... that reminds me, must remove the bit that hides the output :)

agubler:
I think we should add all the ejected config inside a root folder such as `config` similar to `create-react-app`, otherwise it will clutter the root directory.

src/commands/eject.ts
```diff
@@ -142,7 +27,20 @@ function register(options: OptionsHelper): void {
 	});
 }
 
-function run(helper: Helper, args: EjectArgs): Promise<any> {
+function copyFiles(commandName: string, { path, files }: FileCopyConfig): void {
+	if (path && files && files.length > 0) {
```
agubler: are we missing unit test for one of these branches?

src/commands/eject.ts
```diff
@@ -142,7 +27,20 @@ function register(options: OptionsHelper): void {
...
+		files.forEach((fileName) => {
+			const cwd = process.cwd();
```
agubler: does this change every time?

tomdye: fair one

src/commands/eject.ts
tomdye: may be, will check

src/npmInstall.ts
```diff
@@ -1,22 +1,31 @@
...
 			})
 			.on('error', (err: Error) => {
-				spinner.stopAndPersist(red.bold(' failed'));
 				reject(err);
 			});
 	});
 }
+
+export async function installDevDependencies({ devDependencies = {} }: NpmPackage) {
+	return npmInstall(['--save-dev', ...convertToInlineDependencies(devDependencies)]);
```
jdonaghue: @tomdye I like the idea of not munging their `package.json` but what if they want to commit the result of the eject? The result would not be portable. Should we persist a `package.json` within each ejected command perhaps? I don't know what the solution is but it seems like that might be a problem.

agubler: @jdonaghue why would this not be portable? The new dependencies are being added to the `package.json` using the `--save-dev` and `--save` flags depending on the type of dependency.

tomdye: it does get persisted James, it's just done by `--save` and `--save-dev` as you would on the command line yourself.

jdonaghue: @agubler @tomdye ah I missed that, duh its right there in front of me now that you mention it. Sorry, don't mind me :)

tomdye:
@agubler all checks / coverage passed now

PR #110, opened 2017-02-22
Change config helper to use sync and have config per command

tomdye:
- sync not async because `register` command where this could be used is `sync`.
- change API to take a `commandName` to avoid config conflicts between commands.
- checking `ejected` state in `eject` command before ejecting.
- change config helper to expose `get` / `save` functions and not be a class.
- check `ejected` state before loading a command.

Blocks: https://github.com/dojo/cli-build/issues/58
Fixes: https://github.com/dojo/cli/issues/112 #104 #106


src/ConfigurationHelper.ts
```diff
@@ -1,10 +1,29 @@
...
-import { ConfigurationHelper } from './interfaces';
+import { ConfigurationHelper, Config } from './interfaces';
 const pkgDir = require('pkg-dir');
-const packageName = pkgDir.sync(__dirname).split('/').pop();
 
 const appPath = pkgDir.sync(process.cwd());
+const dojoRcPath = join(appPath, '.dojorc');
+
+function writeConfigFile(config: Config) {
+	writeFileSync(dojoRcPath, JSON.stringify(config), { flag: 'wr' });
```
agubler: Can we pretty format the output please?

src/ConfigurationHelper.ts
```diff
@@ -1,10 +1,29 @@
...
+	const configExists = existsSync(dojoRcPath);
+	const config: Config = configExists ? readJsonSync(dojoRcPath) : {};
+	let writeFile = !configExists;
+
+	if (commandName && !config.hasOwnProperty(commandName)) {
+		config[commandName] = {};
+		writeFile = true;
+	}
+
+	writeFile && writeConfigFile(config);
```
agubler: Why are we writing the file when we get the config? Doesn't seem like we'd need to do this until a command calls save?

src/ConfigurationHelper.ts
tomdye: can do.

src/ConfigurationHelper.ts
```diff
@@ -13,42 +32,26 @@ const appPath = pkgDir.sync(process.cwd());
...
-						reject(error);
-						return;
-					}
-					resolve();
-				});
-			});
-		});
+	save(config: Config, commandName: string): void {
+		const dojoRc = getConfigFile();
+		const commmandConfig: Config = dojoRc.hasOwnProperty(commandName) ? dojoRc[commandName] : {};
```
agubler: `const commandConfig: Config = dojoRx[commandName] || {};`

src/ConfigurationHelper.ts
```diff
@@ -13,42 +32,26 @@ const appPath = pkgDir.sync(process.cwd());
...
+
+		Object.assign(commmandConfig, config);
+		Object.assign(dojoRc, { [commandName]: commmandConfig});
```
agubler: Any reason you cannot use `deepAssign` from `@dojo/core`? 

```ts
deepAssign(dojoRc, { [commandName]: commandConfig });
```

Worth testing as I know there was an issue raised against `dojo/core/lang#deepAssign` the other day.

tomdye: `deepAssign` would likely cover the whole thing, ie. i wouldn't need to do two assigns, BUT, I want to ensure i'm not overwriting any other command config so I guess i could just change the first one to `deepAssign`

src/ConfigurationHelper.ts
```diff
@@ -1,10 +1,29 @@
...
-const packageName = pkgDir.sync(__dirname).split('/').pop();
 
 const appPath = pkgDir.sync(process.cwd());
+const dojoRcPath = join(appPath, '.dojorc');
+
+function writeConfigFile(config: Config) {
+	writeFileSync(dojoRcPath, JSON.stringify(config), { flag: 'wr' });
+}
+
+function getConfigFile(commandName?: string): Config {
```
agubler: I don't like this name if it gets the config file or a commands part of the config which it looks like it does.

src/ConfigurationHelper.ts
agubler: Why would you override any other command config with a deep assign?

tomdye:
@agubler this is ready for review now. Fully working and tested.

src/configurationHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+const pkgDir = require('pkg-dir');
+
+const appPath = pkgDir.sync(process.cwd());
+const dojoRcPath = join(appPath, '.dojorc');
+
+function writeConfigFile(config: Config) {
+	writeFileSync(dojoRcPath, JSON.stringify(config, null, 2));
+}
+
+function getConfigFile(commandName?: string): Config {
```
agubler: This should not take the `commandName` imo; It should just return the config file (or `{}` if it doesn't exist)

src/configurationHelper.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	writeConfigFile(dojoRc);
+};
+
+/**
+ * Retrieves the configuration object from the file system
+ *
+ * @returns Promise - an object representation of .dojorc
+ */
+export function get(commandName: string): Config {
```
agubler: This should be 

```ts
    const configFile = getConfigFile();
    return configFile[commandName] || {};
```

src/configurationHelper.ts
tomdye: OK, so the exposed `get` function that calls it will do that bit. Good call, I thought i'd done that already as you commented on it before.

src/configurationHelper.ts
tomdye: Had done this before u commented ;)

PR #120, opened 2017-03-08
Use `@types/yargs`

### LINKED ISSUE #119: Migrate to @types/yargs from typings.json
**Enhancement** 

Migrate to `@types/yargs` from yargs dependency in `typings.json`

agubler:
**Type:** feature

**Description:**

Convert to use @types/yargs.

*Discussion* as we do lose the typings for `Argv` that seem to be typed as `any` in the latest yarg types from @types.

Resolves #119


src/CommandHelper.ts
```diff
@@ -22,7 +22,7 @@ export default class implements CommandHelper {
 	private commandsMap: CommandsMap;
 	private context: any;
 	private configuration: ConfigurationHelper;
-	run(group: string, commandName?: string, args?: yargs.Argv): Promise<any> {
+	run(group: string, commandName?: string, args?: any): Promise<any> {
```
tomdye: can't this become `Argv` ?

agubler: No because what was `Yargs` is now `Argv` and the typing for what was `Argv` is now `any`.

agubler: I did mention this in the description.

tomdye: I see that now.

dylans:
@agubler any reason to not land this after a rebase?

agubler:
@dylans just the fact that the typings don't seem to be quite as good. Also it's a breaking change as we'd need to convert all the commands.

PR #126, opened 2017-03-16
Making option value required, issue #45

### LINKED ISSUE #45: Support multiple groups for a command

rorticus:
**Type:** feature

* [ ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**

Adding some documentation on specifying the right option names, and made option value required, not optional.

Resolves #45

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -38,9 +38,13 @@ export interface Alias {
 	options: AliasOption[];
 }
 
+/**
+ * Represents one alias option. The option name is the full name of the option, and not the abbreviation. For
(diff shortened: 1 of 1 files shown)


tomdye:
Thanks @rorticus

PR #137, opened 2017-03-24
Restricting configuration helpers to only working for a single command.

### LINKED ISSUE #128: A command should have not read/write access to other commands config
**Bug**

**Expected behavior:**

I should only be allowed to read and write to my own portion of `.dojorc`

**Actual behavior:**
```
helper.configuration.get()
```
and
```
helper.configuration.save()
```

can both access any command config in the `.dojorc`

rorticus:
**Type:** bug / feature

**Description:**

Modifies the configuration helper to not allow it to be called with a command name. It is now _always_ called with the group and command of the command being run.

I'm not really convinced on this, as it seems to me that it was a lot of code to change for one simple task, but I think I like the direction it's going. 

Please give me your feedback!

Resolves #128

### DIFF
--- a/src/CommandHelper.ts
+++ b/src/CommandHelper.ts
@@ -1,8 +1,8 @@
-import { CommandsMap } from './command';
-import { CommandHelper, Command, ConfigurationHelper } from './interfaces';
-import Helper from './Helper';
-import configurationHelper from './configurationHelper';
 import * as yargs from 'yargs';
+import { CommandsMap } from './command';
+import { ConfigurationHelperFactory } from './configurationHelper';
+import HelperFactory from './Helper';
+import { CommandHelper, Command } from './interfaces';
 
 function getCommand(commandsMap: CommandsMap, group: string, commandName?: string): Command | undefined {
 	const commandKey = commandName ? `${group}-${commandName}` : group;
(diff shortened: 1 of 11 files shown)


agubler:
@tomdye can we get a review on this please?

Also @rorticus is this a breaking change?

rorticus:
@agubler I'm afraid this is a breaking change.  The helper coming into commands is modified to remove the command name parameters, so

```ts
configurationHelper.save("cli-eject", { someValues });
```

turns into,

```ts
configurationHelper.save({ someValues });
```

agubler:
@rorticus Do you think we could maintain backwards compatibility by continuing to support both `configurationHelper.save("cli-eject", { someValues });` & `configurationHelper.save({ someValues });`?

For the original API we could just change the implementation to ignore the first argument when there are two arguments passed? And mark it as deprecated?

rorticus:
OK @agubler it looks like I was wrong. The original signature is,

```ts
configurationHelper.save({ someValues }, "cli-eject");
```

so it would be backwards compatible. I did however add the old signatures to the new methods and marked them as deprecated in the JSDOC.

dylans:
This one needs a re-review and hopefully should be landed.

PR #143, opened 2017-04-20
Parse dashed command names properly

### LINKED ISSUE #142: Parsing of hyphenated commands
**Bug** 

Currently if you have a command that includes an additional hyphen, the parsing of the command becomes unexpected.  In the sense that you have a command package of `cli-generate-project-json` this becomes `dojo generate-project json` instead of `dojo generate project-json`.

kitsonk:
**Type:** bug

* [X] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**

This PR fixes issues when a command has a dash in it's name.  Commands like `cli-foo-bar-baz` will be parsed as a group of `foo` and a name of `bar-baz`.  Previously, these were being parsed as `foo-bar` and `baz`.

This will mean that group names cannot be dashed.

Resolves #142

### DIFF
--- a/tests/support/dash-names-foo-bar-baz.ts
+++ b/tests/support/dash-names-foo-bar-baz.ts
@@ -1,0 +1,7 @@
+export const description = 'test-description';
+export function register() {
+	return 'registered';
+};
+export function	run() {
+	return new Promise((resolve) => 'ran');
+}
--- a/tests/support/dash-names-foo-bar-baz.ts
+++ b/tests/support/dash-names-foo-bar-baz.ts
@@ -2,6 +2,6 @@ export const description = 'test-description';
 export function register() {
 	return 'registered';
 };
-export function	run() {
+export function run() {
 	return new Promise((resolve) => 'ran');
 }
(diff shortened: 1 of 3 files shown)


tests/support/dash-names-foo-bar-baz.ts
```diff
@@ -0,0 +1,7 @@
+export const description = 'test-description';
+export function register() {
+	return 'registered';
+};
+export function	run() {
```
agubler: nit: is this formatting funky?

tests/unit/command.ts
```diff
@@ -122,6 +125,16 @@ registerSuite({
 			}
 		}
 	},
+	'load of commands parsed correctly': {
+		'setup'() {
+			loader = command.initCommandLoader(testSearchPrefixesDashedNames);
+			commandWrapper = loader('../tests/support/dash-names-foo-bar-baz');
```
agubler: Can you use `getCommandPath`?

tests/support/dash-names-foo-bar-baz.ts
kitsonk: It is...

tests/unit/command.ts
kitsonk: No, because it has `const testGroup` and `const testName`.

agubler: urgh fair enough

PR #149, opened 2017-05-23
Fix: correctly load .dojorc commands

### LINKED ISSUE #148: Unable to load dojorc configuration
**Bug**

The CLI is unable to load `dojorc` configuration for any command, regardless of whether the default command is used or full command.

The `_configurationKey` set here https://github.com/dojo/cli/blob/master/src/configurationHelper.ts#L26 is not being done so correctly.

Using `dojo build` the key is being set to `build` and using `dojo build webpack` the key is being set to `build-build-webpack`. The key used shoud be `build-webpack`.

mwistrand:
**Type:** bug

**Description:**

Updates the cli helper to take a simplified command name (e.g., "webpack" or "intern") instead of a composite of the group and command (e.g., "build-webpack" or "test-intern"), so that when looking up data from a `.dojorc`, the correct key is used (e.g., "build-webpack" instead of just "build" or "build-build-webpack").

Verified with `@dojo/examples/todo-mvc-kitchensink`, using both the `$ dojo build` and `$ dojo test` cli commands. Note that `@dojo/cli-test-intern` currently does not ask for `.dojorc` data, but a quick change to the local source confirmed that these changes will work whenever that update is made.

Resolves #148

### DIFF
--- a/src/CommandHelper.ts
+++ b/src/CommandHelper.ts
@@ -28,7 +28,7 @@ export class SingleCommandHelper implements CommandHelper {
 		const command = getCommand(this._commandsMap, group, commandName);
 		if (command) {
 			const helper = new HelperFactory(this, yargs, this._context, this._configurationFactory);
-			return command.run(helper.sandbox(group, commandName), args);
+			return command.run(helper.sandbox(group, command.name), args);
 		}
 		else {
 			return Promise.reject(new Error('The command does not exist'));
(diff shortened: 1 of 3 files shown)


agubler:
@mwistrand surprised it didn't break any tests but perhaps it was because the test data was actually wrong.

mwistrand:
@ant Resolving #150 might uncover just that.

PR #152, opened 2017-07-24
Adding ability to specify commands at the command line, issue #125

### LINKED ISSUE #125: Add an option to help with testing commands that are in development
**Enhancement** 

There is no good way with cli to load a command that is "in development."  I.e., you have to and link or copy files around in order to actually test the command you're working on.

It might be helpful to pass a path, or series of paths, that cli could scan to look for commands.

```shell
dojo --dev=/path/to/command,/path/to/another/command
```

I think of this is as similar to installing a Chrome extension in development mode, in which it will read a directory as an extension and not require the extension to be bundled.

rorticus:
**Type:** feature

**Description:**

Adds an option to specify one or more explicit commands to load. This is helpful if you have a command that is:

* Not in `node_modules`
* Does't follow the `@dojo/cli` or `dojo-cli` convention
* Is in development

Specify the module ID at runtime,

```shell
dojo --command=/Projects/dojo2/cli-test-intern/dist/umd/index
```

Module group and name will be read from command description if available, if not available, they will try to be extracted from the file name (looks for `cli-(group)-(name)` pattern).

Resolves #125

### DIFF
--- a/src/command.ts
+++ b/src/command.ts
@@ -1,4 +1,5 @@
-import { Command } from './interfaces';
+import {Command} from './interfaces';
+
 const cliui = require('cliui');
 
 export interface CommandWrapper extends Command {
@@ -12,9 +13,9 @@ export type CommandsMap = Map<string, CommandWrapper>;
 /**
  * Function to create a loader instance, this allows the config to be injected
  * @param: searchPrefix A string that tells the command loader how cli commands will be named, ie.
- * 	'@dojo/cli-' is the default meaning commands could be
- * 		- '@dojo/cli-build-webpack'
- * 		- '@dojo/cli-serve-dist'
+ *    '@dojo/cli-' is the default meaning commands could be
+ *        - '@dojo/cli-build-webpack'
+ *        - '@dojo/cli-serve-dist'
  */
 export function initCommandLoader(searchPrefixes: string[]): (path: string) => CommandWrapper {
 	const commandRegExp = new RegExp(`(?:${searchPrefixes.join('|').replace('\/', '\\/')})-([^-]+)-(.+)$`);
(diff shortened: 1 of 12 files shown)


src/config.ts
```diff
@@ -1,19 +1,32 @@
...
+	builtInCommandLocation: string,
+	explicitCommands: string[]
 };
 
+const explicitCommands: string[] = [];
+
+if (yargs.argv && yargs.argv.command) {
+	const paths = Array.isArray(yargs.argv.command) ? yargs.argv.command : [yargs.argv.command];
+
+	paths.forEach((path: string) => {
```
kitsonk: also can't we do...

```ts
const explicitCommands: string[] = yargs.argv && args.argv.command && Array.isArray(yargs.argv.command) ? targs.argv.command : [ yargs.argv.command] || [];
```

doing a `Array.forEach` to push onto another array is really an odd construct... if all else failed, I would have simply spread the array, like `[ ...paths ]` instead of forEach... but even then it can all be simplified.

src/config.ts
```diff
@@ -1,19 +1,32 @@
...
+
 export default {
 	searchPaths: [
 		'node_modules',
 		join(__dirname, '..', '..'),
 		join(packagePath, 'node_modules')
 	],
-	searchPrefixes: [ '@dojo/cli', 'dojo-cli' ],
-	builtInCommandLocation: join(__dirname, '/commands')  // better to be relative to this file (like an import) than link to publish structure
+	searchPrefixes: ['@dojo/cli', 'dojo-cli'],
```
kitsonk: no cuddling...

src/loadCommands.ts
```diff
@@ -44,6 +44,10 @@ export async function enumerateBuiltInCommands (config: Config): Promise <string
 	return globby(builtInCommandParentDirGlob, (<globby.Options> { ignore: '**/*.map' }));
 }
 
+export async function enumerateExplicitCommands (config: Config): Promise <string []> {
```
kitsonk:
```ts
export function enumerateExplicitCommands(config: Config): Promise<string[]> {
```

now we are spacing where we don't space... 😂  also, the `async` annotation is unnecessary unless we are awaiting within the function.  I guess we need to update the style guide about that.

Is there any good reason to defer this API too, since it will always resolve immediately anyways?

src/command.ts
```diff
@@ -24,9 +25,49 @@ export function initCommandLoader(searchPrefixes: string[]): (path: string) => C
 
 		try {
 			const command = convertModuleToCommand(module);
-			const { description, register, run, alias, eject } = command;
+			const {description, register, run, alias, eject} = command;
+			//  derive the group and name from the module directory name, e.g. dojo-cli-group-name
+			const [, group, name] = <string[]> commandRegExp.exec(path);
```
rorticus: Without the cast, it complains that `RegExpExecArray` doesn't have a `Symbol.iterator`

agubler:
I would ask, isn't `grunt link` and `npm link` not good enough for testing commands in development? Perhaps it's not, just wondering.

rorticus:
@kitsonk , responded to your feedback!

rorticus:
@agubler `npm link` might be enough for working with the `@dojo` cli commands, but this might still be useful if you ever want to run a command that isn't provided by `@dojo`, or, maybe one that doesn't even have an npm package because it is part of your local project?

agubler:
@rorticus would `npm link` work any module? I am creating a command called `dojo-cli-my-custom-module`, all I would need to do is `npm link` in my commands root directory and then `npm link dojo-cli-my-custom-module` in my project to test it?

rorticus:
@agubler yeah I think that would work. I wish I could remember the specifics right now, but this ticket was originally put in because testing a certain CLI command was a big pain. There were a couple of different linking steps that had to be done to get the command going, and just being able to specify a path would have greatly simplified things. Maybe that's not an issue anymore.

agubler:
What is the conclusion here? Does the command line for specifying local commands make sense? Or should we just continue to use linking to test local commands? cc @kitsonk @rorticus @matt-gadd

rorticus:
I'm ok with just closing this if linking will do the trick.

kitsonk:
I tend to do linking too and haven't run into any issues.

rorticus:
Closing this. It sounds like linking will do the trick, so this isn't needed.

PR #164, opened 2017-09-29
Make punctuation consistent with other descriptions

umaar:
image


agubler:
Guess we don't have a test for these description messages.

PR #166, opened 2017-10-03
Fix unhandled promise rejections

### LINKED ISSUE #150: Clear `UnhandledPromiseRejectionWarning`s from tests.
**Bug**

Currently there are about a dozen `UnhandledPromiseRejectionWarning`s reported by Node.js when running the unit tests, indicating at the very least something is wrong with the tests themselves, or worse with the actual implementation.

Package Version: next

JamesMilnerUK:
**Type:** Bug

* [X] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**

Resolves #150. It appears the issue was that stub.resolves is used but a promise was being returned back too, creating a double stacked promise. Simply returning a string resolved it.  I've also handled potentially failing promises in the tests themselves, and set process to give a stack trace for any unhandled promises in future.

On another note you still get two 'test error message' from the tests even though it implies they all passed. Any ideas on that? It's the same on master too.

### DIFF
--- a/tests/unit/commands/eject.ts
+++ b/tests/unit/commands/eject.ts
@@ -80,6 +80,7 @@ describe('eject command', () => {
 			group: 'command',
 			name: ''
 		});
+
 		const installedCommandWrapper2 = getCommandWrapperWithConfiguration({
 			group: 'version',
 			name: ''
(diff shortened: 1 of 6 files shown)


tests/unit/commands/eject.ts
```diff
@@ -172,6 +183,8 @@ describe('eject command', () => {
 				assert.isTrue(consoleLogStub.getCall(hintsCall).calledWith(underline('\nhints')), 'should underline hints');
 				assert.isTrue(consoleLogStub.getCall(hintsCall + 1).calledWith(' hint 1'), 'should show hint1');
 				assert.isTrue(consoleLogStub.getCall(hintsCall + 2).calledWith(' hint 2'), 'should show hint2');
+			}).catch(() => {
```
kitsonk: Minor nit, but would prefer we were consistent in starting chains on a new line.

```
})
.catch(() => {
```

JamesMilnerUK: I found a couple other cases in the code base with no new lines, so I fixed those too.

PR #182, opened 2017-12-14
Add cli-create-widget

### LINKED ISSUE #181: Add cli-create-widget
**Enhancement**

Add [cli-create-widget](https://github.com/dojo/cli-create-widget) to the default command set.

Package Version: latest

bitpshr:
**Type:** feature

**Description:**

This PR adds `cli-create-widget` to the default CLI command set.

Resolves #181

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/cli",
-  "version": "0.3.0-pre",
+  "version": "0.3.1-pre",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


agubler:
Do we think that `cli-create-widget` should automatically available when the cli is installed? Seems like you'd only use it in a context of a project, would not having it as a dev dependency of a project work?

tomdye:
Does it scaffold out an entire structure for developing / building said widget the same way that create-app does? if so, then i agree that it should be part of `cli`. On the other hand, if it adds a widget to an existing app, then it should be an installable dev-dependency.

tomdye:
@bitpshr given that this command is aimed at adding `widgets` to an app I feel that it should be added to the `dev-dependencies` of the `package.json` template created by `cli-create-app`. Then this would be available within a newly created app. It is not a command that works stand alone without an app built around it so does not make sense to be shipped as such.

agubler:
As per issue https://github.com/dojo/cli/issues/183, we are going to remove all the bundled commands from the CLI - but regardless I don't think that `cli-create-widget` is truly a "global" command as it will depend on the version of widget-core that is being used in the project.

PR #190, opened 2018-01-03
Bump dependencies to latest. Fix `DOJO_CLI` typing.

tomdye:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [ ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
* [ ] Unit or Functional tests are included in the PR

**Description:**

Bumps dependencies to latest as they were getting outdated.
Switches to the typings shipped with chalk.

Requires: https://github.com/dojo/interfaces/pull/88
Resolves: https://github.com/dojo/cli/issues/188
Resolves: https://github.com/dojo/cli/issues/191

PR #200, opened 2018-01-18
CLI Command list

### LINKED ISSUE #184: Improve command discoverability
**Enhancement**

This kind of continues on from https://github.com/dojo/cli/issues/183, and I think it has been mentioned elsewhere (I think in regards to the create-theme functionality). We need to improve the discoverability of what commands are available for the end user to install. There are multiple ways we can attack this, a few ideas:

* list all of our commands available in `@dojo/cli-*` in when you run `dojo`, regardless of whether they are actually installed (with some different formatting/text, to let the user know they need to install them)

* have a separate command like `dojo list` or similar which shows packages that are available to download from `@dojo/cli-*`

* show some suggestion text that varies every time `dojo` is called like "have you tried @dojo/cli-create-theme?, it lets you scaffold themes quickly"

* if running a command that doesn't exist, ie `dojo build`, ask them if they have installed it and offer them suggestions what package that might be.

tomdye:
**Type:** bug / feature

**Description:**

Adds command discovery ability to the CLI. All commands including those that have not been installed will now show when you run `dojo`.

- bumps dependencies
- moves all interfaces to `src/interfaces`
- shows available commands
- creates detached process to check for available commands on subsequent runs

Screenshots below show `index.js` as I'm running command from build source via `node index.js`.

screenshot 2018-01-19 10 55 19
screenshot 2018-01-19 10 55 31

Resolves #184 
Supersedes: https://github.com/dojo/cli/pull/190


src/index.ts
```diff
@@ -17,15 +18,19 @@ async function init() {
...
+		const mergedCommands = mergeInstalledCommandsWithAvailableCommands(allCommands, availableCommands);
+
+		registerCommands(yargs, mergedCommands.commandsMap, mergedCommands.yargsCommandNames);
 	} catch (err) {
 		console.log(`Commands are not available: ${err}`);
 	}
 }
 
-process.env.DOJO_CLI = true;
+process.env.DOJO_CLI = 'true';
```
umaar: Just curious, what is this for?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
+import * as execa from 'execa';
+import { join } from 'path';
+const spawn: any = require('cross-spawn');
+import { NpmPackageDetails, LoadedCommands,	YargsCommandNames, CommandsMap, CommandWrapper } from './interfaces';
```
umaar: I'm seeing some extra spacing before `YargsCommandNames `

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+export default async function(name: string): Promise<NpmPackageDetails[]> {
+	const conf = new Configstore(name);
+
+	let commands: NpmPackageDetails[] = conf.get('commands') || [];
+	if (!commands.length) {
+		commands = await getLatestCommands(name, conf);
+	} else {
+		const lastUpdated = conf.get('lastUpdated');
+		if (Date.now() - lastUpdated >= ONE_DAY) {
+			spawn(process.execPath, [join(__dirname, 'detachedCheckForNewCommands.js'), JSON.stringify({ name }) ], {
```
umaar: Just to understand, why does it need to be done as a detached process?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+import chalk from 'chalk';
+
+const INITIAL_TIMEOUT = 3000;
+const ONE_DAY = 1000 * 60 * 60 * 24;
+
+export default async function(name: string): Promise<NpmPackageDetails[]> {
+	const conf = new Configstore(name);
+
+	let commands: NpmPackageDetails[] = conf.get('commands') || [];
+	if (!commands.length) {
```
umaar: How do you feel about this rule? Personally I like it and thinks it makes sense, https://eslint.org/docs/rules/no-negated-condition if you agree, shall we follow it here?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+				stdio: 'ignore'
+			}).unref();
+		}
+	}
+
+	return commands;
+}
+
+export async function getLatestCommands(packageName: string, conf: Configstore): Promise<NpmPackageDetails[]> {
+	let commands: NpmPackageDetails[] = [];
```
umaar: How come we're using `let`?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+	commands = await search(INITIAL_TIMEOUT);
+	conf.set('commands', commands);
+	conf.set('lastUpdated', Date.now());
+
+	return commands;
+}
+
+async function search(timeout: number = 0): Promise<NpmPackageDetails[]> {
+	return execa('npm', ['search', '@dojo', 'cli-', '--json', '--searchstaleness', '0'], { timeout }).then(result => {
+		const commands = JSON.parse(result.stdout);
```
umaar: Can we use just `stdout` here (so destructure it in the callback argument) ?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+
+	commands = await search(INITIAL_TIMEOUT);
+	conf.set('commands', commands);
+	conf.set('lastUpdated', Date.now());
+
+	return commands;
+}
+
+async function search(timeout: number = 0): Promise<NpmPackageDetails[]> {
+	return execa('npm', ['search', '@dojo', 'cli-', '--json', '--searchstaleness', '0'], { timeout }).then(result => {
```
umaar: Can we use `await` instead of a `then` callback?

src/installableCommands.ts
umaar: As this is coming from something external, shouldn't a `JSON.parse` always be wrapped in a `try/catch`?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+		const installCommand = `npm i ${command.name}`;
+		const commandWrapper: CommandWrapper = {
+			name,
+			group,
+			path: installCommand,
+			description: command.description,
+			register: () => {},
+			run: () => {
+				console.log(`\nTo install this command run ${chalk.green(installCommand)}\n`);
+				return Promise.resolve();
```
umaar: Why do we need to promise.resolve for a sync function?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+	availableCommands.forEach((command) => {
+		const [ , group, name ] = regEx.exec(command.name) as string[];
+		const compositeKey = `${group}-${name}`;
+		const installCommand = `npm i ${command.name}`;
+		const commandWrapper: CommandWrapper = {
+			name,
+			group,
+			path: installCommand,
+			description: command.description,
+			register: () => {},
```
umaar: What is this empty function for?

src/installableCommands.ts
```diff
@@ -0,0 +1,111 @@
...
+			run: () => {
+				console.log(`\nTo install this command run ${chalk.green(installCommand)}\n`);
+				return Promise.resolve();
+			}
+		};
+
+		if (!isEjected(group, name)) {
```
umaar: If it's not long to explain, can you explain what is happening here for the user? ....But maybe I should just study this closer 😄

tests/unit/installableCommands.ts
```diff
@@ -0,0 +1,153 @@
...
+		sandbox.stub(console, 'log');
+	});
+
+	afterEach(() => {
+		sandbox.restore();
+		mockModule.destroy();
+	});
+
+	it('should check for installable commands if the config store is empty', () => {
+		return moduleUnderTest.default('testName').then((commands: NpmPackageDetails[]) => {
```
umaar: Can we use async/await for all the tests please? Instead of then callbacks

tests/unit/installableCommands.ts
```diff
@@ -0,0 +1,153 @@
...
+			assert.equal(commands.length, 1);
+			assert.equal(commands[0].name, '@dojo/cli-test');
+		});
+	});
+
+	it('creates installable command prompts', () => {
+		const { commandsMap, yargsCommandNames } = moduleUnderTest.createInstallableCommandPrompts([ testCommandDetails, testCommandDetails2 ]);
+		assert.equal(commandsMap.size, 3);
+		assert.isTrue(commandsMap.has('test'));
```
umaar: I don't think it's necessary for assert this one, because we already do `commandsMap.get('test').name`?

Also because if it fails it can leads to errors like expected false to be true or something which is hard to debug!

tests/unit/installableCommands.ts
```diff
@@ -0,0 +1,153 @@
...
+		assert.equal(commandsMap.get('test').name, 'command');
```
umaar: How about doing a deep equal on the object returned from `commandsMap.get('test')` instead?

tests/unit/installableCommands.ts
```diff
@@ -0,0 +1,153 @@
...
+		commandsMap.set('installed', { name: 'installed-command' });
+		yargsCommandNames.set('installed', new Set());
+		const installedCommands = {
+			commandsMap,
+			yargsCommandNames
+		};
+		const mergedCommands = moduleUnderTest.mergeInstalledCommandsWithAvailableCommands(installedCommands, [ testCommandDetails ]);
+
+		assert.equal(mergedCommands.commandsMap.size, 3);
+		assert.isTrue(mergedCommands.commandsMap.has('test'));
```
umaar: Similar to the note earlier, maybe using `equals` is better e.g. `mergedCommands.commandsMap.get` instead of `has`? Or how about if we use `isTrue`, we try and add a helpful message?

src/installableCommands.ts
tomdye: because the `interface` requires a function

src/commands/eject.ts
```diff
@@ -1,15 +1,15 @@
 import { copySync } from 'fs-extra';
 import { basename, isAbsolute, resolve } from 'path';
 import { Argv } from 'yargs';
-import { green, underline, yellow } from 'chalk';
```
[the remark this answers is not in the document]
tomdye: The new typings for `chalk` bundled with `chalk` error if it is not done this way.

src/index.ts
tomdye: This is for the `cli-build-webpack` command to know if it's running in the cli or not

src/installableCommands.ts
tomdye: agreed, it's just like this because i wrote the other case afterwards, will fix

src/installableCommands.ts
tomdye: because the cli will have ran and finished before the search comes back and we do not want to delay the cli message / function every time. This is the same way that `yeoman/update-notifier` works

src/installableCommands.ts
tomdye: my mistake / refactor

src/installableCommands.ts
tomdye: sure

src/installableCommands.ts
tomdye: required via `commandwrapper` interface

src/installableCommands.ts
tomdye: if the user has ejected the command then we do not want to show it as an available command. we do the same in the code that loads all installed commands

tests/unit/installableCommands.ts
tomdye: I'm personally not a fan of async/await in the tests as I feel it makes them harder to debug

src/index.ts
agubler: It's not actually used by the new `cli-build-app` command if that helps.

tests/unit/installableCommands.ts
tomdye: not required as the only property i have passed it in this test is `name`

tests/unit/installableCommands.ts
tomdye: I avoided doing this as i know the `commandWrapper` interface as been met due to the face that the code compiles, otherwise it's a large object with functions etc... to replicate and i thought it overkill

tomdye:
re. the color changes, it is something that we have thought about, worth raising an issue for as there is a solution out there for colouring in the description.
Looks like the command name colour itself is a bit of a hack so unlikely to go down that route.

re. the general output, again this should be addressed in an issue, the purpose of this PR is simply to show the available commands in the same format as we currently show installed ones.

tomdye:
@umaar most comments addressed

src/index.ts
tomdye: so this can in fact be removed once we deprecate `cli-build-webpack`

src/installableCommands.ts
tomdye: Because it's a requirement of the interface

src/installableCommands.ts
```diff
@@ -0,0 +1,114 @@
...
+	const commands = await search(INITIAL_TIMEOUT);
+	if (commands && commands.length) {
+		conf.set('commands', commands);
+		conf.set('lastUpdated', Date.now());
+	}
+	return commands || [];
+}
+
+async function search(timeout: number = 0): Promise<NpmPackageDetails[] | undefined> {
+	const { stdout } = await execa('npm', ['search', '@dojo', 'cli-', '--json', '--searchstaleness', '0'], { timeout });
```
umaar: Since this is going over the network, does it need to wrapped in a try/catch? If it errors, is it caught + handled appropriately higher up?

PR #208, opened 2018-02-12
Remove npm dep by removing `david`

### LINKED ISSUE #206: Remove `npm` dependency by removing `david`
**Enhancement** 

The `npm` dependency we have in `cli` causes problems when reinstalling `node_modules` etc and we no longer require `david` as we are using `npm search` elsewhere to collect available version numbers.

We should remove this dependency and change the version command to use the `npm search` results.

tomdye:
**Type:** feature

**Description:**

New `version` data is gathered from `npm search` rather than `david`.

screenshot 2018-02-14 13 39 18

Resolves #206


src/commands/version.ts
```diff
@@ -39,12 +39,15 @@ export interface VersionArgs extends Argv {
...
-type DavidDependencies = {
-	[dependencyName: string]: {
-		stable: string;
-		latest: string;
-	};
-};
+async function getLatestCommandVersions(): Promise<NpmPackageDetails[]> {
+	const packagePath = pkgDir.sync(__dirname);
+	const packageJsonFilePath = join(packagePath, 'package.json');
+	const packageJson = <any>require(packageJsonFilePath);
```
agubler: can we do `const packageJson: any = require('')` or at least `const packageJson = require('') as any;`?

src/commands/version.ts
```diff
@@ -53,38 +56,26 @@ type DavidDependencies = {
...
+
+	const latestCommands = await getLatestCommandVersions();
+	const latestVersions: VersionsMap = latestCommands.reduce((versions: VersionsMap, { name, version }) => {
+		versions[name] = version;
+		return versions;
+	}, {});
+
+	return moduleVersions.map(({ name, version, group }) => {
+		const latestVersion = latestVersions[name];
+		const canBeUpdated = version !== latestVersion;
```
agubler: Is this going to be wrong if you have a later version but it's not latest (i.e. you are using `next`)?

tomdye: David did the same as this, always warns if you're not on the `latest` version.

agubler: Does that means it's correct or expected?

tomdye: We can raise an issue to change it to use a semver comparison at a later date if thats important? Atm I have simply replicated what it did previously.

agubler: Yeah I think it's something we can think about in the future

PR #214, opened 2018-03-11
Ensure exit code is non zero when command fails

### LINKED ISSUE #209: A broken test builde returns with exit code zero
**Enhancement**

When building tests with `dojo build -m test` the return code to the operating system is zero even if the build process fails.

Package Version: 0.5.0

**Expected behavior:**

A non zero return code will possibly be more appropriate and ensure that the build breaks a CI build chain.

agubler:
**Type:** bug / feature

**Description:**

Exit with non zero code when running a command fails.

Resolves #209

### DIFF
--- a/tests/unit/registerCommands.ts
+++ b/tests/unit/registerCommands.ts
@@ -22,6 +22,7 @@ let yargsStub: any;
 let defaultRegisterStub: SinonStub;
 let defaultRunStub: SinonStub;
 let consoleErrorStub: SinonStub;
+let processExitStub: SinonStub;
 const errorMessage = 'test error message';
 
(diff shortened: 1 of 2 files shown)


tests/unit/registerCommands.ts
```diff
@@ -203,26 +209,18 @@ registerSuite('registerCommands', {
...
 					return {
-						beforeEach() {
-							processExitStub = stub(process, 'exit');
-						},
-						afterEach() {
-							processExitStub.restore();
-						},
-
 						tests: {
 							async 'Should not exit process if no status code is returned'() {
```
agubler: I should change the name of this test.

agubler: and I forgot before I merged :doh:

PR #220, opened 2018-03-23
Add global option to save config in the dojorc

### LINKED ISSUE #211: Add `--save-args` flag to write new command defaults to dojorc
**Enhancement** 

To enhance the dojorc defaults we could add a `--save-args` flag that writes new defaults to `dojorc`.
For example `dojo build -m dev --save-args` would save `mode: 'dev'` to the `dojorc` so that the next time you ran `dojo build` without args it would pick up the dev mode argument.

agubler:
**Type:** feature

**Description:**

Adds a global option to the CLI that will save the options passed to the `.dojorc` that will be used when running the command with no options on subsequents runs.

Resolves #211

### DIFF
--- a/src/registerCommands.ts
+++ b/src/registerCommands.ts
@@ -23,22 +23,19 @@ function reportError(error: CommandError) {
 	process.exit(exitCode);
 }
 
-function userSetOption(option: string, parsed: any) {
-	function searchForOption(option: string) {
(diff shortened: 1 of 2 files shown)


JamesMilnerUK:
This should be brought up to speed with master : https://github.com/dojo/cli/pull/255/files

PR #229, opened 2018-04-02
Override yargs help output with custom help renderer

### LINKED ISSUE #227: Help shows options for the default command when regardless of what command is requested
**Bug**

When help is called on a specific command but it's not the "default" command for the group, the help currently still only shows the default command options and puts all the requested commands options at the bottom in the main "global" options section.

Also the help displays duplicates option choices.

1__anthony_red-sky____development_test-beds_dojo-cli

Package Version: latest (0.6.2)

### LINKED ISSUE #228: Create custom help output for cli and commands
**Enhancement**

Currently the CLI uses the default help output from `yargs` for the cli help output. This has been fine up to now, but it is very basic looking and hard to customise.

We should consider managing the CLI help explicitly and write a custom help output that is targeted for our exact CLI architecture. 

This would enable use to create some nice output for the CLI help.

agubler:
**Type:** feature

**Description:**

Handles help output explicitly for the CLI and commands with a custom help renderer. Also needed some significant changes in the way that commands are registered:

* Normalise the previous `CommandMap` and `YargsCommandNames` to a single `GroupMap` so that they are not managed over two objects
* Added extra optional property on the command API for `global`, defaults to `false`
* Added extra options on the internal `CommandWrapper` - `global`, `installed`, `default`
* `default` is set on the first registered command for a group (same as before)
* `installed` is set to false for commands found from `npm`
* Removed support for command `aliases` as no commands currently use it and it complicates the help renderer.

Temporarily hard coded `create-app` to be a global command as it's the "dojo" global external command.

### Main CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

### Group CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

### Command CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

### Installable command CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

Resolves #227  
Resolves #228


agubler:
Am completely open to suggestions to the formatting and colours and also if I've missed anything.

agubler:
Needs more test coverage!

src/CommandHelper.ts
```diff
@@ -19,12 +15,12 @@ export type RenderFilesConfig = {
  * allowing commands to call one another. Provides 'run' and 'exists' functions
  */
 export class SingleCommandHelper implements CommandHelper {
-	private _commandsMap: CommandsMap;
+	private _groupMap: GroupMap;
```
rorticus: Funny that we're renaming these to "commands" in the README and changing them FROM commands in the code 😂

src/command.ts
```diff
@@ -26,6 +25,8 @@ export function initCommandLoader(searchPrefixes: string[]): (path: string) => C
 				alias,
 				description,
 				register,
+				installed: true,
+				global: group === 'create' && name === 'app' ? true : global,
```
rorticus: Maybe we should move this into a utility or something so that the next time we need to force a global like this we don't have to modify the logic in here?

src/CommandHelper.ts
agubler: Yeah it’s because we already have a commands map that is the value of the group map. Traditionally it’s easier to understand these things as different (as they really are) when it comes to code. Open to other name suggestions though.

src/command.ts
agubler: This is a totally temporary thing, this PR introduces an optional global property on the API that commands can mark whether that are global and then we can remove this logic completely

agubler:
@tomdye I think for now the padding works but I defo want to look at using cli-table for the layout.

tests/unit/validation.ts
```diff
@@ -0,0 +1,111 @@
...
+				assert.fail('Should throw a validation error');
+			} catch (err) {
+				assert.strictEqual(
+					err.message,
+					"\n\u001b[1m\u001b[31mError(s):\u001b[39m\u001b[22m\n  Required option '\u001b[91mfoo\u001b[39m' not provided\n  Required option '\u001b[91mbar\u001b[39m' not provided"
+				);
+			}
+		});
+
+		it('Should not return validation error for when option provided', () => {
```
dylans: "Should not return validation error for when option provided" might want to make this clearer?

agubler:
**Updated CLI output with some extra colour and different ascii art for dojo**

### Main CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

### Group CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

### Command CLI help output

1__anthony_red-sky____development_test-beds_dojo-cli

agubler:
@tomdye @rorticus Have changed the output slightly if you could take another look.

rorticus:
wow that's fancy :)

agubler:
Art courtesy of @matt-gadd 😄

Issue #205, opened 2018-02-09
README Review


edhager:
- The TOC does not include all of the headings.
- Under "The CLI has the following in-built groups:", put each dojo command in a bullet so it is easier to scan the document and find that list.
  - dojo create - provides scaffolding for new Dojo 2 projects. 
  - dojo eject - allows users to configure and run command instead of the cli. 
  - dojo version - provides information on the versions of installed commands and the cli itself.
- There's a warning about ejecting but no real explanation of how it works.  We should describe where the configuration files are placed and give example of ejecting, modifying a config and running the tool.  
- Is this really true about ejecting?  "This action is one-way and you cannot go back to having the tools managed by the cli."  If I eject, can't I still run the cli commands?  It just will not use the ejected config files right?

agubler:
> - Under "The CLI has the following in-built groups:", put each dojo command in a bullet so it is easier to scan the document and find that list.
>  - dojo create - provides scaffolding for new Dojo 2 projects. 
>  - dojo eject - allows users to configure and run command instead of the cli. 
>  - dojo version - provides information on the versions of installed commands and the cli itself.

We should remove `dojo create` as `cli-create-app` is no longer bundled with the `CLI`

> - There's a warning about ejecting but no real explanation of how it works.  We should describe where the configuration files are placed and give example of ejecting, modifying a config and running the tool.  

👍 

> - Is this really true about ejecting?  "This action is one-way and you cannot go back to having the tools managed by the cli."  If I eject, can't I still run the cli commands?  It just will not use the ejected config files right?

It is true, once ejected, the commands will not be runnable through the `dojo cli`, `ejecting` should (and does I think for each of the commands) provide the command needed to run after ejection.

edhager:
> dojo build and dojo test are not installed by default with @dojo/cli. To use them, you must install them separately, e.g. `npm i @dojo/cli-build-webpack` and `npm i @dojo/cli-test-intern` These two groups are not included by default to allow different versions of these groups to be installed per project.

@dojo/cli-build-webpack should be @dojo/cli-build-app right?

Issue #204, opened 2018-02-09
README review


jason0x43:
## Notes

The subsections below refer to sections in the readme.

### (General notes)

Standardize on CLI or cli. Also, are we using "cli" as a name or a description? (So is it "the cli" or just "cli" everywhere?)

> The CLI is the officially supported way...

Maybe _This package is the officially supported way..._ or _`cli` is the officially supported way..._

### Why use

> It is designed to save you time, by promoting a standardised workflow, and automating away more mundane boilerplate tasks.

Remove the commas.

> ...multiple tools such as Webpack, Intern and tslint, you can...

[webpack](https://webpack.js.org/branding/#the-name) should be lower-cased.

> When you eject, all the configuration and build dependencies of the included tools will be moved into your project. If you are adept at configuring these tools, then you can now do so without the cli using its defaults.

Maybe _When you eject, all the configuration and build dependencies of the included tools will be moved into your project. The development process can then be tailored to the specific needs of your project._

### Usage

> Getting the cli

Remove this.

> ...then please re-install and make sure the installation runs without errors.

This might be better worded as "verify that the installation runs" or "check that the installation runs".

### Features

This section is kind of confusing. It might be clearer if "group" and "command" were renamed to "command" and "subcommand", which is a more common terminology. (Even if the group isn't technically a "command", it looks and feels like a command to the user.) Then drop the first few paragraphs and just start at "The CLI has the following built-in options:". Also, prefer "built-in" to "in-built".

> The CLI has the following in-built groups:

_The CLI has the following built-in commands_

In the paragraph listing the built-in commands, put them in a list (one per line or bulleted).

> To use them, you must install them separately, e.g.
>
> npm i @dojo/cli-build-webpack

This paragraph has a weird line break in it, and there should be a period after `npm I @dojo/cli-test-intern`.

### How do I contribute?

Would this be better as "How can I contribute?"

> An additional npm script to run prettier (with write set to true) against all src and test project files is available by running:

Maybe better worded as _You format all `src` and `test` project files by running:_

> In order to build the project run grunt dev or grunt dist.

It would be nice to add npm scripts for these tasks to the package.json and reference those instead. (This goes for the testing commands, too).

> Test cases MUST be written using Intern using the Object test interface and Assert assertion interface.

_using the "object" test interface and "assert" assertion interface_

agubler:
If we want to update the "How do I contribute" section (I think these suggestions are good), we would need to do it across all the packages!

agubler:
[...]

PR #232, opened 2018-04-06
README updates

rorticus:
A collection of README updates from https://github.com/dojo/cli/issues/204 and https://github.com/dojo/cli/issues/205. 

Note that I didn't do too much with the _How can I contribute?_ section because from reading the comments on the issue, that might be a bigger change across all the repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ You can list all your global npm dependencies by running:
 
 `npm list -g --depth=0`
 
-If you don't see `@dojo/cli` in the list of global dependencies, verify that the installation runs with out errors.
+If you don't see `@dojo/cli` in the list of global dependencies, verify that the installation runs without errors.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,40 +61,39 @@ You can list all your global npm dependencies by running:
 
 `npm list -g --depth=0`
 
-If you don't see `@dojo/cli` in the list of global dependencies, then please re-install and make sure the installation runs without errors.
+If you don't see `@dojo/cli` in the list of global dependencies, verify that the installation runs with out errors.
```
dylans: it should be without, not with out.

README.md
```diff
@@ -149,7 +152,7 @@ In order to build the project run `grunt dev` or `grunt dist`.
 
 ### Testing
 
-Test cases MUST be written using [Intern](https://theintern.github.io) using the Object test interface and Assert assertion interface.
+Test cases MUST be written using [Intern](https://theintern.github.io) using the "object" test interface and "assert" assertion interface.
```
dylans: Should these be in quotes, or backticks?

rorticus: Not actually sure... issue https://github.com/dojo/cli/issues/204 uses double quotes, and the PR from you that I clobbered doesn't use quotes at all. I changed it to backticks just now and it looks OK to me 😄

PR #236, opened 2018-05-18
Adding support for reading/writing config from package.json

rorticus:
**Type:** feature

**Description:**

Adding the ability to read from `.dojorc` configuration from `package.json`. If the `dojo` key exists in your `package.json`...

```json
{ 
  "dojo" : {
    "build-app": {
    }
  }
}
```
And `.dojorc` does **not** exist, the config will be read from `package.json`. When you try to write config, and it's stored in `package.json`, you get a confirmation.

```text
You are using a "dojo" configuration in your package.json. Saving the current settings will update your package.json. Continue? [ (N)o / (Y)OLO ]:
```

Additionally, you can now pass a command name into the `get()` method in order to read configuration that belongs to another command. Note that this applies to **read** only. You still cannot write to another command's config.

All of this is to support future work on `@dojo/cli-create-theme` to read a list of themeable elements that will be stored in `package.json`.


rorticus:
Closing in favor of https://github.com/dojo/cli/pull/260

PR #243, opened 2018-07-15
Add validate command and validate commands before running with provided schema

### LINKED ISSUE #241: .dojorc validation
**Enhancement**

Some CLI tools that rely on configs will do validation on those configs to ensure they are correct. For example Webpack does this:

screenshot_20180710_141137

I'm not entirely sure how this would be achieved, so I thought I'd open that up for discussion. Happy to look into implementing once there is consensus on the best approach.

**Code**

I would imagine the interface would be something like this:

`dojo validate`

or  for example 

`dojo build` 

Throws an error if  `.dojorc` doesn't match a specified schema

JamesMilnerUK:
Feature

**Description:**

Resolves #241 

This should allow commands to be validated against a validation schema that is provided in the command source (at the moment it's named `rc-schema.json` but we could go with something else). If no config is provided validation will skip for the command. 

screenshot_20180823_000019

screenshot_20180822_235902


agubler:
@JamesMilnerUK I’ve been thinking about the command validation and general API, which are different to how you’ve currently implemented it....

I’ve been thinking that we should provide a json schema validation function via the helper and the command API for validate should be a function instead of a boolean. 

In the `validate` function commands would/could use the provided json schema validation function from the CLI. This would make the validation flexible as a command can run completely bespoke validation or extra validation on top of the basic schema validation. It would also mean that you should have to worry about paths for the command.

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+		);
+
+		validateCommands(commands, helperFactory);
+	});
+}
+
+export default {
+	name: '',
+	group: 'validate',
+	description: 'validate your .dojorc build for installed commands',
```
agubler: what does `build` mean here? Shouldn't it be `.dojorc configuration file` or something?

JamesMilnerUK: Yeah you're right, not sure what I was getting out :sweat_smile:

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+	commands.forEach((commandMap, group) => {
+		toValidate = [...commandMap.values()].reduce((toValidate, command) => {
+			if (typeof command.validate === 'function') {
+				toValidate.add(command);
+			}
+			return toValidate;
+		}, toValidate);
+	});
+	return toValidate;
+}
```
agubler: Could perhaps make this a little clearer... not hugely important the end result is (or should be) the same:

```ts
const commandsToValidate = commands.reduce((commandWrappers, command) => {
	const groups = [ ...command.values() ];
	const filtered = groups.filter((group) => (typeof group.validate === 'function'));
	commandWrappers = [ ...commandWrappers, ...filtered ];
	return commandWrappers;
}, [] as CommandWrapper);

return new Set<CommandWrapper>(commandsToValidate);
```

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+		!validation.silentSuccess && logSchemaSuccess(commandKey);
+	}
+
+	return noMismatches;
+}
+
+function validateCommands(commands: Map<string, Map<string, CommandWrapper>>, helper: HelperFactory) {
+	const config = getConfigFile();
+	const noConfig = config === undefined;
+	const emptyConfig = typeof config === 'object' && Object.keys(config).length === 0;
```
agubler: Could this just be `!!config && !Object.keys(config).length`?

agubler: In fact you could inline them:

```ts
if (!config) {
    // stuff
} else if (!Object.keys(config)) {
    // stuff
}

File: src/registerCommands.ts [hunk 4]
```diff
@@ -134,12 +137,15 @@ function registerGroups(yargs: Argv, helper: HelperFactory, groupName: string, c
...
-
-				const args = getOptions(
-					aliases,
-					helper.sandbox(groupName, defaultCommand.name).configuration.get(),
-					argv
-				);
+				const config = helper.sandbox(groupName, defaultCommand.name).configuration.get();
+				const args = getOptions(aliases, config, argv);
+
+				if (typeof defaultCommand.validate === 'function') {
```
agubler: Can you just check it exists? It's only allowed to be a function on the interface?

src/registerCommands.ts
```diff
@@ -177,7 +192,14 @@ function registerCommands(yargs: Argv, helper: HelperFactory, groupName: string,
 export default function(yargs: Argv, groupMap: GroupMap): void {
 	const helperContext = {};
 	const commandHelper = new CommandHelper(groupMap, helperContext, configurationHelperFactory);
-	const helperFactory = new HelperFactory(commandHelper, yargs, helperContext, configurationHelperFactory);
+	const validateHelper = { validate: builtInCommandValidation }; // Provide the default validation helper
```
agubler: How come this is an object, to allow for possible extension in the future?

tests/support/validate/corruptSchema.json
```diff
@@ -0,0 +1,2 @@
+{
```
agubler: Lol I was about to comment on this, but then noticed the name of the file!

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+export function logNoConfig() {
+	console.log(yellow(`No config has been detected`));
+}
+
+export function logValidateFunctionFailed(error: Error) {
+	console.log(red(`The validation function for this command threw an error: ${error}`));
+}
+
+export function logEmptyConfig() {
+	console.log(yellow(`A config was found, but it has no properties`));
```
agubler: Don't need to use template literals here

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+const { red, green, yellow } = chalk;
+
+export interface ValidateArgs extends Argv {}
+
+type ValidationErrors = string[];
+
+function register(options: OptionsHelper): void {}
+
+export function logNoConfig() {
+	console.log(yellow(`No config has been detected`));
```
agubler: Don't need to use template literals here

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+import configurationHelperFactory, { getConfigFile } from '../configurationHelper';
+import { Validator } from 'jsonschema';
+import CommandHelper from '../CommandHelper';
+import HelperFactory from '../Helper';
+
+const { red, green, yellow } = chalk;
+
+export interface ValidateArgs extends Argv {}
+
+type ValidationErrors = string[];
```
agubler: Do you really need a type for this? It looks like it is only used once (and it's just a basic type!)

src/commands/validate.ts
agubler: and in other places too!

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export function logNoValidatableCommands() {
+	console.log(green(`There were no commands to validate against`));
+}
+
+export function getValidationErrors(commandKey: string, commandConfig: any, commandSchema: any): ValidationErrors {
+	const validator = new Validator();
+	const result = validator.validate(commandConfig, commandSchema);
+
+	const errors = result.errors.map((err: any) => {
```
agubler: Can just return without the assignment if you want :)

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+				toValidate.add(command);
+			}
+			return toValidate;
+		}, toValidate);
+	});
+	return toValidate;
+}
+
+export function builtInCommandValidation(validation: ValidationWrapper): boolean {
+	const commandKey = validation.commandGroup + '-' + validation.commandName; // group and name are required properties
```
agubler: could use template literals here

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+			if (typeof command.validate === 'function') {
+				toValidate.add(command);
+			}
+			return toValidate;
+		}, toValidate);
+	});
+	return toValidate;
+}
+
+export function builtInCommandValidation(validation: ValidationWrapper): boolean {
```
agubler: Might be worthwhile destructuring validation 
```ts
const { commandGroup, commandName, commandSchema, commandConfig, silientSuccess } = validation;
```

src/commands/validate.ts
```diff
@@ -0,0 +1,157 @@
...
+	const commandKey = validation.commandGroup + '-' + validation.commandName; // group and name are required properties
+
+	if (validation.commandConfig === undefined) {
+		logSchemaErrors(`.dojorc config does not have the top level command property '${commandKey}'`);
+		return false;
+	}
+
+	const mismatches = getValidationErrors(commandKey, validation.commandConfig, validation.commandSchema);
+	let noMismatches = true;
```
agubler: could just be `noMismatches = !mismatches.length;`, in fact you could just return that and save a variable (the name `noMismatches` always takes me ages to process for some reason!)

agubler:
General note, do you think that we need to report successful validation for each command? Especially when running the command (as the user doesn't actually need to know that it is validated before running)... for the validate built in I agree we should return a general, "All you commands are configured correctly" or what have you.

src/commands/validate.ts
JamesMilnerUK: Yeah not a bad shout

src/commands/validate.ts
JamesMilnerUK: :+1:

src/commands/validate.ts
JamesMilnerUK: Yeah trying to weigh this one up, I do like the original explicitness but because we're using Typescript I guess we can assume type correctness to some degree. I do like the terseness of your suggestion though so I'm going to go with it!

src/registerCommands.ts
JamesMilnerUK: Yeah true, although if someone overrode this for some bizarre reason it would break the command.

src/registerCommands.ts
JamesMilnerUK: Yeah, could be a function, but as you said if we ever want to extend the helper it's nice to not break the API

src/commands/validate.ts
JamesMilnerUK: I'm not sure reduce works on `Map`s unfortunately.  There is probably a way to simplify this though. Let me have a think.

agubler: Sorry that might have needed a `.values()` but otherwise the same

JamesMilnerUK: I think this is probably the clearest:
```
let toValidate = new Set<CommandWrapper>();
	commandMaps.forEach((commandMap) => {
		[...commandMap.values()].forEach((command) => {
			if (typeof command.validate === 'function') {
				toValidate.add(command);
			}
		});
	});
	return toValidate;
```

JamesMilnerUK:
@agubler as it stands each command is responsible for deciding if `silentSuccess` should be set. If that's set to `true` then nothing will log on success regardless of how the validate command is run. Because we invert the control to the command to setup validate, we would have to say it's just best practice to set that to `false` for installable commands, if that makes sense.

src/commands/validate.ts
```diff
@@ -0,0 +1,158 @@
...
+		}
+	});
+
+	if (noMismatches) {
+		logConfigValidateSuccess();
+	}
+}
+
+async function run(helper: Helper, args: ValidateArgs): Promise<any> {
+	return loadExternalCommands().then(async (commands) => {
```
rorticus: Why is this `async`?

JamesMilnerUK: The main reason was because eject uses async here: https://github.com/dojo/cli/blob/0ec95762448c2b6528fe9e674f7e86a45a00e738/src/commands/eject.ts#L53

src/commands/validate.ts
```diff
@@ -0,0 +1,158 @@
...
+
+export function logConfigValidateSuccess() {
+	console.log(green('There were no issues with your config!'));
+}
+
+export function logNoValidatableCommands() {
+	console.log(green('There were no commands to validate against'));
+}
+
+export function getValidationErrors(commandKey: string, commandConfig: any, commandSchema: any): string[] {
```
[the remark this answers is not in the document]
JamesMilnerUK: `commandConfig` is all the installable command configuration schema. I'm not a massive fan of having them hardcoded because they will probably change often and new commands will be made available. `commandSchema` would be even more complicated/error prone to define!

src/commands/validate.ts
agubler: I think @rorticus meant, why using `async` keywords rather than why is it returning a promise (as it's not using `await` anywhere)

agubler: I think @rorticus meant, why using `async` keywords rather than why is it returning a promise (as it's not using `await` anywhere)

JamesMilnerUK: Yeah let's just remove it

PR #253, opened 2018-08-15
Show an error when no package.json is found for builtin commands requiring it

### LINKED ISSUE #252: Running CLI and commands without package.json throws errors
**Enhancement** 

When `dojo` or associated commands are run in a directory without a `package.json` file, the CLI throws errors such as `Path must be a string. Received null`

Package Version: 3.0.0

**Code**

```dojo``` or ```dojo init``` in any directory without a `package.json`

**Expected behavior:**

The CLI to either behave as best as possible or give a meaningful error message if not.

**Actual behavior:**

Unhelpful end user errors

JamesMilnerUK:
**Type:** bug / feature

**Description:**

Resolves #252

** Screenshot **

screenshot_20180815_175643

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,14 +4,16 @@ import registerCommands from './registerCommands';
 import { join } from 'path';
 import commandLoader from './allCommands';
 import installableCommands, { mergeInstalledCommandsWithAvailableCommands } from './installableCommands';
+import { existsSync } from 'fs';
+import chalk from 'chalk';
 const pkgDir = require('pkg-dir');
 
 export async function init() {
-	try {
-		const packagePath = pkgDir.sync(__dirname);
-		const packageJsonFilePath = join(packagePath, 'package.json');
-		const packageJson = <any>require(packageJsonFilePath);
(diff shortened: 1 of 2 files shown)


agubler:
I'm still not sure about adding the warning whilst we have a common command that should and does work without a `package.json`. I can imagine that will be confusing.

JamesMilnerUK:
@agubler we could perhaps make it more granular, i.e. when it actually tries to access `package.json`?

agubler:
@JamesMilnerUK That's often done in the actual commands I think - which the CLI doesn't know about.

JamesMilnerUK:
So for init for example it happens here:
https://github.com/dojo/cli/blob/master/src/commands/init.ts#L11
Version:
https://github.com/dojo/cli/blob/master/src/commands/version.ts#L43
I think this will also be a big offender:
https://github.com/dojo/cli/blob/master/src/config.ts#L3

It would be good to have a unified way of trying to access the package.json and then erroring if it fails, even if it's just for the builtins. What do you think @ant?

PR #255, opened 2018-09-04
Add support to save command line options to the dojorc

### LINKED ISSUE #211: Add `--save-args` flag to write new command defaults to dojorc

JamesLMilner:
**Type:** feature

**Description:**

@agubler already wrote this but it is not in line with latest master. This should solve that.

Resolves #211


JamesMilnerUK:
screenshot_20181004_214845

I changed this to support showing `--save` in the help feedback

PR #256, opened 2018-09-04
Add support for creating directories that don't exist on renderFiles

### LINKED ISSUE #247: writeRenderedFile() should create the destination directories
**Enhancement** 

Both `@dojo/cli-create-app` and `@dojo/cli-create-widget` have to manually create the directories they are rendering into before calling `helper.command.renderFiles()`. This should be done automatically by `writeRenderedFiles()`.

Package Version: 3.0.0-alpha.1

JamesMilnerUK:
**Type:** Feature

**Description:**

Resolves #247

### DIFF
--- a/src/template.ts
+++ b/src/template.ts
@@ -1,6 +1,5 @@
 import { renderFile } from 'ejs';
-import { dirname } from 'path';
-import { writeFile, existsSync, mkdirSync } from 'fs-extra';
+import { writeFile, ensureDir } from 'fs-extra';
 import chalk from 'chalk';
 
 export function ejsRender(source: string, replacements: Object): Promise<string> {
@@ -16,8 +15,7 @@ export function ejsRender(source: string, replacements: Object): Promise<string>
 
 export function writeRenderedFile(str: string, destination: string): Promise<void> {
 	return new Promise<void>((resolve, reject) => {
-		console.log('?????? \n');
-		ensureDirectoryExistence(destination);
+		ensureDir(destination);
 		writeFile(destination, str, (err?: Error) => {
 			if (err) {
 				reject(err);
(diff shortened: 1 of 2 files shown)


src/template.ts
```diff
@@ -15,6 +16,8 @@ export function ejsRender(source: string, replacements: Object): Promise<string>
 
 export function writeRenderedFile(str: string, destination: string): Promise<void> {
 	return new Promise<void>((resolve, reject) => {
+		console.log('?????? \n');
```
bryanforbes: This should be removed

src/template.ts
```diff
@@ -15,6 +16,8 @@ export function ejsRender(source: string, replacements: Object): Promise<string>
...
+		ensureDirectoryExistence(destination);
```
bryanforbes: Why not use [`fs-extra.ensureDir()`](https://github.com/jprichardson/node-fs-extra/blob/master/docs/ensureDir.md)?

src/template.ts
```diff
@@ -15,6 +15,7 @@ export function ejsRender(source: string, replacements: Object): Promise<string>
 
 export function writeRenderedFile(str: string, destination: string): Promise<void> {
 	return new Promise<void>((resolve, reject) => {
+		ensureDir(destination);
```
bryanforbes: `ensureDir()` is an asynchronous function and can either take a callback (like `writeFile()`) or returns a promise.

JamesMilnerUK: Ahh yes.

src/template.ts
```diff
@@ -1,26 +1,22 @@
...
 
 export function writeRenderedFile(str: string, destination: string): Promise<void> {
-	return new Promise<void>((resolve, reject) => {
-		writeFile(destination, str, (err?: Error) => {
-			if (err) {
-				reject(err);
-			}
-			resolve();
-		});
+	return ensureDir(destination).then(() => {
```
bryanforbes: Just for sanity's sake, can you make sure that both of these functions return a promise in the version of `fs-extra` we have pinned?

JamesMilnerUK: For sure!

JamesMilnerUK: @bryanforbes looks like they do:

screenshot_20180904_171506

JamesMilnerUK: and writeFile does also:
screenshot_20180904_180337

PR #259, opened 2018-09-05
Use libnpmsearch rather than calling npm with execa

### LINKED ISSUE #251: Running older versions of npm with execa prevents retriving installable commands as JSON
**Bug**

When using older versions of npm such as `3.5.2` (this is what you get when you install nodejs with Ubuntu by default), `execa` will fail to get the JSON response as intended when passing in the `search` command with necessary arguments. 

screenshot_20180815_115024

From reasoning this seems to be because the `--json` argument is being ignored which causes it to fail.

**Code**

Try:

```javascript
const execa = require("execa");

async function test() {
    console.log( 
        await execa('npm', ['search', '@dojo', 'cli-', '--json',  '--searchstaleness=0'])
    );
}

test();
```

Using a older version of NPM. 

**Expected behavior:**

JSON response that doesn't throw an error in the dojo cli `search` function.

**Actual behavior:**

```Invalid response from npm search```

### LINKED ISSUE #258: Replace execa calls to npm with the official npm libary
**Enhancement** 

At the moment we use `execa` to call out to npm to do a search. This is somewhat convoluted and arguably error prone.

Package Version: 3.0.1

**Expected behavior:**

Not to make calls out to shell calling npm

**Actual behavior:**

We make calls out to shell to call npm

### PR DESCRIPTION
**Pull request by JamesMilnerUK**
**Type:** bug / feature

**Description:**

Resolves #258
Also potentially resolves #251

### CODE REVIEW

File: src/installableCommands.ts [hunk 1]
```diff
@@ -30,25 +29,28 @@ export default async function(name: string): Promise<NpmPackageDetails[]> {
...
-			timeout
-		});
-		const commands = JSON.parse(stdout);
-		return commands.filter(({ name }: NpmPackageDetails) => {
-			return /^@dojo\/cli-/.test(name);
-		});
-	} catch (e) {
-		console.error('Invalid response from npm search');
+		const results = await search('@dojo/cli-');
+		const filteredResults = results
```
bryanforbes: Just `return results`. There's no reason to store it.

src/installableCommands.ts
```diff
@@ -30,25 +29,28 @@ export default async function(name: string): Promise<NpmPackageDetails[]> {
...
+			.filter((result) => {
+				return result.scope === 'dojo' && result.name !== '@dojo/cli';
+			})
+			.map((result) => {
```
bryanforbes: Change this to `.map(({ name, version, description }) => {` and the next line can be removed.

JamesMilnerUK: Good idea :+1:

tests/unit/installableCommands.ts
```diff
@@ -17,7 +17,8 @@ describe('installableCommands', () => {
 	let mockConfigurationHelper: any;
 	let configHelperGetStub: sinon.SinonStub;
 	let mockCrossSpawn: any;
-	let mockExeca: any;
+	let mockLibNpmSearch: any;
```
agubler: can we type this, so if the API changes then the tests will fail to compile?

JamesMilnerUK: As we mentioned on slack, I'm not sure there's an easy way to achieve that as it's the mock as opposed to the actual library. I think it'd be easy to leave it (as we don't currently try and type any other mocks).

src/libnpmsearch.d.ts
```diff
@@ -0,0 +1,29 @@
...
+		scope: string;
+		description: string;
+		version: string;
+		keywords: string[];
+		date: Date | null;
+		maintainers: NpmPackageMaintainers[] | null;
+	}
+
+	interface NpmSearchOpts {
+		limit: number;
```
bryanforbes: Should these properties be optional?

tests/unit/installableCommands.ts
```diff
@@ -72,31 +73,68 @@ describe('installableCommands', () => {
...
 			assert.isTrue(mockConfigStoreGet.calledWith('commands'), 'checks for stored commands');
-			assert.isTrue(mockExeca.ctor.notCalled, 'does not call npm search');
+			assert.isTrue(mockLibNpmSearch.ctor.notCalled, 'does not call npm search');
 		});
 	});
 
-	it('filters out @dojo/cli from search results', () => {
-		mockExeca.ctor.resolves({
-			stdout: `[{"name": "dojo-cli-helper"}, {"name": "@dojo/cli"}, {"name": "@dojo/cli-test"}]`
+	it('filters out none dojo scoped commands from search results', () => {
```
bryanforbes: s/none/non-/

src/libnpmsearch.d.ts
agubler: They should I believe

PR #260, opened 2018-09-05
Read/save package.json dojo config and preserve indent for all configs

JamesMilnerUK:
**Type:** feature

**Description:**

Description:

Adding the ability to read from .dojorc configuration from package.json. If the dojo key exists in your package.json...

```json
{ 
  "dojo" : {
    "build-app": {
    }
  }
}
```

And .dojorc does not exist, the config will be read from package.json. When you try to write config, and it's stored in package.json, you get a confirmation.

You are using a "dojo" configuration in your package.json. Saving the current settings will update your package.json. Continue? [ (N)o / (Y)es ]:

Additionally, you can now pass a command name into the get() method in order to read configuration that belongs to another command. Note that this applies to read only. You still cannot write to another command's config.

All of this is to support future work on @dojo/cli-create-theme to read a list of themeable elements that will be stored in package.json.

### DIFF
--- a/src/configurationHelper.ts
+++ b/src/configurationHelper.ts
@@ -8,16 +8,10 @@ const pkgDir = require('pkg-dir');
 
 const appPath = pkgDir.sync(process.cwd());
 let dojoRcPath: string;
+let packageJsonPath: string;
 if (appPath) {
 	dojoRcPath = join(appPath, '.dojorc');
-}
-
-const packageJsonPath = join(appPath, 'package.json');
-let hasPackageConfig = false;
-if (existsSync(packageJsonPath)) {
-	const { dojo } = JSON.parse(readFileSync(packageJsonPath, 'utf8'));
-
-	hasPackageConfig = typeof dojo === 'object';
+	packageJsonPath = join(appPath, 'package.json');
 }
 
 function readPackageConfig() {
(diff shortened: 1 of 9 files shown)


src/configurationHelper.ts
```diff
@@ -1,21 +1,55 @@
...
+			)
+		);
+	}
+
+	if (canWriteToPackageJson) {
+		const packageJson = JSON.parse(readFileSync(packageJsonPath, 'utf8'));
+
+		packageJson.dojo = config;
+
+		writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
```
bryanforbes: This doesn't preserve the user's tab style in their `package.json` like `npm` does. Have we looked into preserving it?

src/configurationHelper.ts
```diff
@@ -1,21 +1,55 @@
...
+	}
 }
 
 function writeConfigFile(config: Config) {
 	writeFileSync(dojoRcPath, JSON.stringify(config, null, 2));
 }
 
+function dojoRcExists() {
+	return !!dojoRcPath && existsSync(dojoRcPath);
```
bryanforbes: Use `Boolean()` instead of `!!`

src/configurationHelper.ts
JamesMilnerUK: Looks like we already use it

PR #268, opened 2018-10-22
Merge command groups instead of overriding

### LINKED ISSUE #197: Built in commands and installed commands under same group will overwrite one another
**Bug** 

If there is a built in command with a group named `create` and an installed command with group `create` they will overwrite each other rather than merging when the command list is created.

This is because a spread operator is used to merge commands at the top level within the cli [here](https://github.com/dojo/cli/blob/master/src/allCommands.ts#L46).
`commands.commandsMap = new Map([...installedCommands.commandsMap, ...builtInCommands.commandsMap]);`

KaneFreeman:
**Type:** bug

**Description:**
When the same command group is built in and installed, the commands should be merged instead of the built in command group overriding the installed.

Resolves #197


agubler:
Can we get a unit test for this?

KaneFreeman:
@agubler I have moved the combining code into a separate function and built a unit test for that function.

PR #267, opened 2018-10-07
Better handle calling help on non existent commands

JamesMilnerUK:
**Type:** bug

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)
* [x] Unit or Functional tests are included in the PR

**Description:**

At the moment the CLI returns the help template on non existent commands. This PR fixes that.

Resolves #266 

screenshot_20181015_115521

PR #269, opened 2018-10-29
Handle non existant commands when passed by the user with help flag

### LINKED ISSUE #266: Dojo CLI will return help for none existent commands
**Bug** 

Currently dojo cli will log the help template for none existent commands

Package Version: 3.0.2

**Code**

`dojo nonsense -h`

**Expected behavior:**

Message saying this command does not exist

**Actual behavior:**

screenshot_20181004_220129

JamesMilnerUK:
**Type:** bug 

* [X] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

I had a first go at this with #267 but after speaking with @agubler felt it was better to display the actual full help with a notification that the command where help is being requested does not exist.

Resolves #266


src/help.ts
```diff
@@ -185,6 +190,18 @@ ${formatHelpOutput(groupMap, isNpmCommand)}
 `;
 }
 
+function formatMissingCommandHelp(groupMap: GroupMap, fullCommand: string) {
+	return `${formatHeader()}
+
+${chalk.bold(
+		chalk.red(`Specified command '`) +
```
bryanforbes: I would use `"` instead of backticks here. I had to do a double-take to figure out what was going on.

src/help.ts
```diff
@@ -185,6 +190,18 @@ ${formatHelpOutput(groupMap, isNpmCommand)}
...
+			chalk.redBright(fullCommand) +
+			chalk.red(`' does not exist, please see available commands below.`)
```
bryanforbes: I would do the same thing with the backticks here.

PR #270, opened 2018-10-30
Add LoggingHelper

### LINKED ISSUE #140: Logging to console/stdout in a consistent fashion
**Enhancement** 

Related to #133

In order to have a consistent look and feel, we should have an abstraction API for logging to the console for CLI commands.  The API should provide a set of semantic functions that allow for a consistent look and feel (but also make it easy to maintain, test, etc.).

For example, we could have something like:

```typescript
interface CLILogger {
  commandStart(name: string, ...args: any[]): void;
  commandEnd(): void;
  action(name: string, ...args: any[]): void;
  processStart(name: string, ...args: any[]): void;
  processEnd(): void;
  verbose(message: string, ...args: any[]): void;
}
```

Also, in lots of places we log to `console.log()` but in some cases we have to write to `stdout`.  I would say for CLI commands, we should generally manage writing to any `WritableStream` (but defaulting to `process.stdout`).

bryanforbes:
**Type:** feature

**Description:**

Resolves #140

### DIFF
--- a/src/template.ts
+++ b/src/template.ts
@@ -29,7 +29,7 @@ export default async function(
 	replacements: Object,
 	logging: LoggingHelper
 ): Promise<void> {
-	logging.info(chalk.bold(' create ') + destination);
+	logging.info(chalk.green.bold(' create ') + destination);
 	const str = await ejsRender(source, replacements);
 	await writeRenderedFile(str, destination);
 }
--- a/src/template.ts
+++ b/src/template.ts
@@ -2,6 +2,7 @@ import { renderFile } from 'ejs';
 import { writeFile, ensureDir } from 'fs-extra';
 import * as path from 'path';
 import chalk from 'chalk';
+import { LoggingHelper } from './interfaces';
 
 export function ejsRender(source: string, replacements: Object): Promise<string> {
 	return new Promise<string>((resolve, reject) => {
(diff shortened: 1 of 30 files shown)


src/template.ts
```diff
@@ -22,8 +23,13 @@ export function writeRenderedFile(str: string, destination: string): Promise<voi
...
 
-export default async function(source: string, destination: string, replacements: Object): Promise<void> {
-	console.info(chalk.green.bold(' create ') + destination);
+export default async function(
+	source: string,
+	destination: string,
+	replacements: Object,
+	logging: LoggingHelper
+): Promise<void> {
+	logging.info(chalk.bold(' create ') + destination);
```
KaneFreeman: This change makes the entire line green instead of just the word create. This has a downstream impact on [dojo/cli-create-theme](https://github.com/dojo/cli-create-theme). The theme.ts file is created by this `template` function, but other files are also created directly from the [dojo/cli-create-theme](https://github.com/dojo/cli-create-theme) code, and it has mimicked the styling here of: green create, file name.

That code would need to be updated to match this change.

See https://github.com/dojo/cli-create-theme/pull/43

bryanforbes: Thanks for pointing this out. I've updated the code to match the old behavior.

agubler:
Closing as the changes are very outdated now with significant conflicts. It would be easier to re-raise the changes if a logging helper is something that we still want the cli to offer.

PR #271, opened 2018-11-09
Make validate asynchronous

### LINKED ISSUE #272: Type mistmatch between return types of ValidateHelper and validate
**Bug**

There is currently a mismatch between the a commands `.validate` method and the ValidateHelper. `.validate` returns type of boolean but the ValidateHelper returns a Promise, meaning you can't use the helper in a command. Making `.validate` async would fix this issue.

**Code**

```javascript
	validate(helper: Helper) {
		let schema;
		try {
			schema = JSON.parse(readFileSync(path.join(__dirname, 'schema.json')).toString());
		} catch (error) {
			return Promise.reject(Error('The dojorc schema for cli-build-app could not be read: ' + error));
		}
		return helper.validation.validate({
			commandGroup: command.group as string,
			commandName: command.name,
			commandSchema: schema,
			commandConfig: helper.configuration.get(),
			silentSuccess: true
		});
	}
```

**Expected behavior:**

You should be able to use the provided `ValidateHelper` in the `.validate` function.

**Actual behavior:**

`ValidateHelper` is async which stops it from being useable.

screenshot_20181130_113710

JamesMilnerUK:
**Type:**  feature

* [X] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

At the moment validate expects a boolean return type but the ValidateHelper returns a Promise, meaning you can't use the helper in a command.

Resolves #272

### DIFF
--- a/src/commands/validate.ts
+++ b/src/commands/validate.ts
@@ -94,7 +94,7 @@ export function builtInCommandValidation(validation: ValidationWrapper): Promise
 	});
 }
 
-function validateCommands(commands: Map<string, Map<string, CommandWrapper>>, helper: HelperFactory) {
+async function validateCommands(commands: Map<string, Map<string, CommandWrapper>>, helper: HelperFactory) {
 	const config = getConfig();
 
 	const noConfig = config === undefined;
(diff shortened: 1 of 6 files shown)


agubler:
Please don't merge this until we have created the v4 branch.

src/interfaces.d.ts
```diff
@@ -42,7 +42,7 @@ export type ValidationWrapper = {
 };
 
 export interface ValidateHelper {
-	validate(validateOpts: ValidationWrapper): Promise<any>;
+	validate(validateOpts: ValidationWrapper): Promise<boolean>;
```
JamesMilnerUK: @agubler @matt-gadd does this look like the interface we want? I was thinking about `Promise<any>` but I can't think if we'd want to return anything other than true or false back from a validate call. What do you think?

src/commands/validate.ts
```diff
@@ -114,21 +114,21 @@ function validateCommands(commands: Map<string, Map<string, CommandWrapper>>, he
...
-	let noMismatches = true;
-
-	[...toValidate].forEach((command) => {
-		try {
-			const valid = !!command.validate && command.validate(helper.sandbox(command.group, command.name));
-			noMismatches = valid && noMismatches;
-		} catch (error) {
-			logValidateFunctionFailed(error);
-			noMismatches = false;
+	return Promise.all(
```
agubler: I find this really hard to follow, do you think we could make it clearer? Also I don't think you want to await in the `map`, as you just want to `await` the resulting array, right?

src/interfaces.d.ts
agubler: Seems like a `boolean` result for validate fits

PR #274, opened 2019-01-07
Add ajv validation

JamesMilnerUK:
**Type:** feature

**Description:**

Uses ajv instead of `jsonschema`

screenshot_20190128_224651


src/commands/validate.ts
```diff
@@ -73,7 +211,7 @@ export function builtInCommandValidation(validation: ValidationWrapper): Promise
 		const commandKey = `${commandGroup}-${commandName}`; // group and name are required properties
 
 		if (validation.commandConfig === undefined) {
-			logSchemaErrors(`.dojorc config does not have the top level command property '${commandKey}'`);
+			logSchemaTopLevelError(commandKey);
```
JamesMilnerUK: I've just realised if there is no config at all (i.e. no `.dojorc` or one in the `package.json`) this branch will trigger. The question is do we want that behaviour? If people have no `.dojorc` we probably should be trying to validate at all.

agubler:
@JamesMilnerUK What is the status of this?

JamesMilnerUK:
@agubler I think it's 90% there. I think as I left a comment above it tries to validate when no config is found which is probably undesirable. There's a pretty comprehensive test suite there, but might benefit from some more smoke testing with some real project configs.

agubler:
@JamesMilnerUK yup a smoke test for commands with validations would help (with a `.dojorc` and also a `package.json` dojo config), also agree that it should not validate when no configuration is found.

PR #281, opened 2019-05-17
Perf: Only glob for commands of that type when a command is passed in the cli args

matt-gadd:
**Type:** feature

**Description:**
This is a simple perf fix for when running a single command. At the moment we glob for all commands even when a command is passed in the cli args. This just narrows it to only glob for that command as a prefix, ie running:
```
dojo test
```
previously would result in:
```
/node_modules/@dojo/cli-*
```
and now results in:
```
/node_modules/@dojo/cli-test-*
```


matt-gadd:
Superseded by https://github.com/dojo/cli/pull/283

PR #283, opened 2019-05-20
Performance improvements

### LINKED ISSUE #275: Startup time of dojo CLI is high
**Enhancement**

At the moment the startup time for running `dojo` is around 2000+ milliseconds:

screenshot_20190204_224842

It looks like the majority of the time is spent loading modules.

Here's the profile as a ZIP:
[CPU-20190204T224612.cpuprofile.zip](https://github.com/dojo/cli/files/2830087/CPU-20190204T224612.cpuprofile.zip)

Package Version: 5.0.1-pre

**Code**

`dojo`

**Expected behavior:**

The CLI loads snappily

**Actual behavior:**

There is a noticeable delay

JamesMilnerUK:
**Type:** bug / feature

**Description:**

This PR aims to do two things:

* Only glob and load commands that match the command that is passed in (rather than loading all)
* When using `dojo`, `dojo --help`, `dojo -h` do not actually load the commands in question as they will never be used 

Supersedes #282 #281 

Resolves #275 

Before:

Screenshot_20190520_171817

After:
Screenshot_20190520_171724


tests/unit/loadCommands.ts
```diff
@@ -162,6 +163,24 @@ registerSuite('loadCommands', {
...
 				}
 			}
+		},
+		'specific command': {
+			afterEach() {
+				mockModule.destroy();
+				testSandbox.restore();
+			},
+			tests: {
+				async 'should not load commands that arent specified group'() {
```
maier49: "arent in"?

tests/unit/loadCommands.ts
```diff
@@ -162,6 +163,24 @@ registerSuite('loadCommands', {
...
+					process.argv = ['node', 'dojo.js', 'bar'];
+					const installedPaths = await enumInstalledCommands(goodConfig);
+					assert.isTrue(installedPaths.length === 0);
+				},
+				async 'should not load commands that are in specified group'() {
```
maier49: "should load"?

src/loadCommands.ts
```diff
@@ -15,8 +15,10 @@ export function isEjected(groupName: string, command: string): boolean {
  * @returns {Promise<string []>} the paths of all installed commands
  */
 export async function enumerateInstalledCommands(config: CliConfig): Promise<string[]> {
+	const [, , command] = process.argv;
```
maier49: Will `command` be defined here if I just run `dojo build`?

JamesMilnerUK: It will indeed as it's `['node', 'path/to/script/', 'build']

tests/unit/loadCommands.ts
JamesMilnerUK: Good spot

src/loadCommands.ts
matt-gadd: this is my fault I think. the variable should be named `group` as it narrows by group not command.

JamesMilnerUK: I'll update

PR #287, opened 2019-08-13
Support `config` command line option to specify a custom dojorc configuration file

### LINKED ISSUE #286: Support for custom `.dojorc` configuration file
Add support for specifying a custom dojorc file name when using the dojo/cli.

agubler:
**Type:** feature

**Description:**

Adds support for specifying a custom configuration file that will be used instead of the default `.dojorc` file. This option will be available for all commands.

Resolves #286


src/registerCommands.ts
```diff
@@ -128,6 +128,7 @@ function registerGroups(yargs: Argv, helper: HelperFactory, groupName: string, c
 				.option('h', {
 					alias: 'help'
 				})
+				.option('config', { default: '.dojorc', type: 'string', description: 'The dojorc config file' })
```
mwistrand: Since this is contained within a `command` call that has its second argument set to `false`, this won't be discoverable with `dojo --help`, correct? If so, is that intentional?

agubler: It's is discoverable in help by adding it explicitly in the help codez.

PR #297, opened 2019-09-24
.dojorc extends support with @speedy/json-extends

### LINKED ISSUE #295: Support extending dojorc configuration
**Enhancement**

In Dojo 6 we added support for using a custom dojorc file with the CLI. This opened many possibilities for creating multiple configurations with tweaked options simply by using the `--dojorc` option. 

Unfortunately this requires all common configuration to be duplicated across each dojorc file, we should support extending dojorc files.

JamesMilnerUK:
**Type:**  feature

**Description:**

The aim is to add a `extends` property to `.dojorc` which will allow for extension for a `.dojorc` from another previously defined `.dojorc` configuration. 

This PR aims to:

* Allow for multiple nested extension, i.e. a `.dojorc` file can extend another `.dojorc` file which can extend another and so forth indefinitely.
* Provide deep 'merging' of configuration objects (i.e. configuration property values that are also objects will be recursively merged). This will only work for Objects and not Arrays or other types (we have agree this isn't such a great idea as per https://github.com/Microsoft/TypeScript/issues/20110).

Resolves #295

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "typings.json"
   ],
   "dependencies": {
-    "@speedy/json-extends": "^1.2.0",
+    "@speedy/json-extends": "1.2.0",
     "ajv": "6.6.2",
     "chalk": "2.4.1",
     "configstore": "3.1.2",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "typings.json"
   ],
   "dependencies": {
+    "@speedy/json-extends": "^1.2.0",
     "ajv": "6.6.2",
     "chalk": "2.4.1",
     "configstore": "3.1.2",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "typings.json"
   ],
   "dependencies": {
+    "@speedy/json-extends": "^1.2.0",
```
agubler: can we pin the dep please?

src/loadCommands.ts
```diff
@@ -72,7 +73,7 @@ export async function loadCommands(paths: string[], load: (path: string) => Comm
 					}
 				}
 			} catch (error) {
-				error.message = `Failed to load module ${path}\nNested error: ${error.message}`;
+				error.message = `${chalk.red(`Failed to load module ${path}`)}\n\nNested error:\n ${error.message}`;
```
agubler: Is this change related?

PR #300, opened 2020-02-04
added force command

### LINKED ISSUE #298: Support force with the schema validation
**Enhancement**

We should add support for a user to be able to "force" run a command even if the schema validation fails.

devpaul:
**Type:** feature

**Description:**

Adds the `--force` parameter allowing for commands to run even when their validation fails. Still prints validation errors and issues, but allows for the user to continue.

Resolves #298


agubler:
@devpaul looks like there is a compilation error on ci

```
src/template.ts(8,36): error TS2345: Argument of type '(err: Error, str?: string | undefined) => void' is not assignable to parameter of type 'Options | undefined'.
  Type '(err: Error, str?: string | undefined) => void' has no properties in common with type 'Options'.
```

src/help.ts
```diff
@@ -164,6 +164,7 @@ function formatCommandOptions(commandWrapper: CommandWrapper, isDefaultCommand =
 	}
 
 	register(formatOption, null as any);
+	formatOption('force', { type: 'boolean', description: 'continue running commands even if validation fails' });
```
agubler: Could we capitalise `continue` in the description?

src/registerCommands.ts
```diff
@@ -143,7 +144,7 @@ function registerGroups(yargs: Argv, helper: HelperFactory, groupName: string, c
 
 				if (typeof defaultCommand.validate === 'function') {
 					const valid = await defaultCommand.validate(helper.sandbox(groupName, defaultCommand.name));
-					if (!valid) {
+					if (!valid && !args.force) {
```
agubler: As we are running the validation still, which means that it will output to the console... Will this look strange for commands that have output (most of them), also it will probably mostly be ignored by users. As using `force` is mostly going to be used after the user has tried to run the command but failed, do you think we should run the validation at all?

devpaul: I prefer to run the validation and output it to the console. If for no other reason it'll be logged in CI for future reference which can be very helpful in tracking back why something did or did not happen.

It ends up looking like this

image

It should display something alarming as `--force` is non-standard. If one of my teammates accidentally checked in a `.dojorc` with the "force" option I'd want alarm bells to go off.

agubler: @devpaul Does putting `force` in the `.dojorc` work? I don't think that it should.

matt-gadd: we should not support `force` in the `.dojorc`, that seems like a terrible idea.

devpaul: Yes. Putting `force` in `.dojorc` works.

Cli arguments and `.dojorc` are merged here https://github.com/dojo/cli/blob/master/src/registerCommands.ts#L142.

We can use `argv` instead of `args` to check for `--force`.

I still think we should display validation errors when `--force` has been used. Warning about validation errors is the same as https://github.com/dojo/cli-build-app/issues/296 where we're warning a user always that the server shouldn't be used in production.

agubler: Okay, we need to make it so that adding `force` to the dojorc is ignored.

devpaul: Done.Now it's kinda ignored.... created bug #302 😂

src/registerCommands.ts
```diff
@@ -183,7 +188,7 @@ function registerCommands(yargs: Argv, helper: HelperFactory, groupName: string,
 
 				if (typeof command.validate === 'function') {
 					const valid = await command.validate(helper.sandbox(groupName, command.name));
-					if (!valid) {
+					if (!valid && !args.force) {
```
agubler: Should this be `argv` too?

devpaul: Good catch. I added a note about that being a gotcha in the architecture documentation and still I did it 😢

PR #303, opened 2020-02-06
Unify execution

devpaul:
**Type:** cleanup

**Description:**

The code to execute a command's validation and run phases exists in two places: once for the default command and once for other commands. It's too easy to make a change to how a command is executed in one spot and not the other.

This PR unifies command validation and execution for both default and other commands.


agubler:
@devpaul I think this looks reasonable, just got a conflict that needs resolving.

